PR #48, opened 2014-06-13
windows clone updates

quintes:
I had some issues pulling the code on win7, so i have updated the git-clone-others.bat and the URL in the README.md

### DIFF
--- a/script/git-clone-others.bat
+++ b/script/git-clone-others.bat
@@ -6,7 +6,7 @@ call :initializeWorkingDirAndScriptDir
 
 rem set startDateTime=`date +%s`
 
-set gitUrlPrefix=git@github.com:droolsjbpm/
+set gitUrlPrefix=https://github.com/droolsjbpm
 rem TODO dynamic gitUrlPrefix detection does not work on mac
 rem cd $scriptDir
 rem gitUrlPrefix=`git remote -v | grep --regex "^origin.*(fetch)$"`
@@ -25,7 +25,7 @@ for /F %%r in ('type %scriptDir%\repository-list.txt') do (
         echo ===============================================================================
         echo Repository: %%r
         echo ===============================================================================
-        call git clone %gitUrlPrefix%\%%r.git %%r
+        call git clone %gitUrlPrefix%/%%r.git %%r
         if "%ERRORLEVEL%" NEQ "0" (
             echo git clone failed
             goto:eof
--- a/README.md
+++ b/README.md
@@ -48,8 +48,9 @@ If you don't want to contribute to this project and you know git and maven, thes
 
 * To build all repositories:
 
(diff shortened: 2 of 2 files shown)


psiroky:
Trying to close/clean-up some old PRs...I think we have two options here:

(1) fix the `git-clone-others.bat` to automatically recognize if it should use ssh or git/https for cloning. This is already being done in the .sh script, so it would just be matter of porting that to the .bat file (not sure how hard/easy would that be

(2) remove the automatic git url prefix recognition from `git-clone-others.sh` and do what @ge0ffrey suggested: ask users whether they want to use ssh (for commiters) or git/https (anonymous access)

We should also mention this in the documentation, so users can pick from ssh or git/https when they are cloning the initial repo (this one).

ge0ffrey:
I like both options. They are both a step forward from the current reality.

psiroky:
I've been briefly looking into this again and it seems to me that adding this kind of logic into .bat script is lot of work. 

I would, however, like to keep the logic in the .sh file because it really is useful. I was thinking about creating PowerShell scripts (instead of plain .bat scripts) where it would be much easier to implement advanced stuff. @mbiarnes, @ge0ffrey  WDYT? Would you be ok with replacing the .bat scripts with PowerShell scripts?

ge0ffrey:
Do powershell scripts work on all modern desktop windows versions? So Windows 7 and every version higher? Do Windows developers recognize the extension as something that is executable on Windows?

A lot of contributors use windows, we don't want to alienate them.

psiroky:
Yes, the powershell is installed on all modern Windows machines (7+). Although on the desktop ones it may be a bit hidden (e.g. the standard command line is more visible). At least that what I encountered few years. I don't really use Windows so my knowledge is limited here.

But since the current version of scripts does not really work as expected (anonymous access is not working), I think these scripts would be a step forward. What I could also do is to create the powershell scripts and then wrap those in .bat files (just calling the powershell script from the .bat).  This should make it easier.

mbiarnes:
+1 linux use .sh, windows powershell (ex .bat) - what about MAC ?

psiroky:
I don't have access to MAC machine to test it, but since it is UNIX based system I expect the .sh files to work there the same as on Linux.

script/git-clone-others.bat
```diff
@@ -6,7 +6,7 @@ call :initializeWorkingDirAndScriptDir
 
 rem set startDateTime=`date +%s`
 
-set gitUrlPrefix=git@github.com:droolsjbpm/
+set gitUrlPrefix=https://github.com/droolsjbpm
```
psiroky: HTTPS is now stable and even recommend by github (https://help.github.com/articles/which-remote-url-should-i-use/).

psiroky:
I've finally merged the change in the `.bat` file as it makes things a bit more easier (anonymous access). In the long term we should consider switching from batch scripts to PowerShell.

PR #49, opened 2014-07-20
Added LogUtils class to contain LogType and LogOrder classes

emilianoandre:
Added prettytime dependency to use in the Logs screen in console

### DIFF
--- a/kie-parent-with-dependencies/pom.xml
+++ b/kie-parent-with-dependencies/pom.xml
@@ -751,7 +751,13 @@
         <artifactId>jetty-runner</artifactId>
         <version>${version.org.mortbay.jetty.runner}</version>
       </dependency>
-
+    
+      <dependency>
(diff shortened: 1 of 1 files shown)


psiroky:
This PR is pretty old. Do we still need to add this dependency?

psiroky:
@emilianoandre @Salaboy do we still need this changes in?

ge0ffrey:
@psiroky I wouldn't be surprised there's an accompanying PR on one of the jbpm repo's, with the actual feature. I don't see a link though.

But despite that, +1 to reject this PR rather than let it rot here. It's a shame that whatever feature was contributed wasn't picked up (aka sponsored) by someone, but it's almost a year, so now it is simply too late.

psiroky:
I could not find a related PR in any of other repos, so closing this one. Please re-open if the changes are still needed.

PR #58, opened 2015-01-19
Make dashbuilder-bom import later than jboss-integration-platform-bom

ryanzhang:
Bug 1183481 - dashbuilder-bom import should be lower priority than jboss-integration-platform-bom import.

In kie-parent-with-dependencies pom.xml, the dashbulder-bom import has the higher priority than jboss-integration-platform-bom. However dashbuilder-bom actually will introduce a dependency on jboss-integration-platform-bom.
This will cause a problem that when dashbuilder-bom is older than jboss-integration-platform-bom, then it will force droolsjbpm project use an older integration-platform-bom. And dashbuilder is outside droolsjbpm project group.
This case happened when I build 6.1.0 ER4. For some reason, dashbuilder-0.2.0.CR2 can't be rebuilt and the latest droolsjbpm still need to use old dashbuilder. 

So the request here is to make dashbuilder-bom define after jboss-integratioin-platform-bom import.

### DIFF
--- a/kie-parent-with-dependencies/pom.xml
+++ b/kie-parent-with-dependencies/pom.xml
@@ -147,13 +147,6 @@
         <scope>import</scope>
       </dependency>
       <dependency>
-        <groupId>org.dashbuilder</groupId>
-        <artifactId>dashbuilder-bom</artifactId>
-        <type>pom</type>
-        <version>${version.org.dashbuilder}</version>
-        <scope>import</scope>
-      </dependency>
-      <dependency>
         <groupId>org.jboss.errai</groupId>
         <artifactId>errai-config</artifactId>
         <version>${version.org.jboss.errai}</version>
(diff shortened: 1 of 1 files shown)


psiroky:
@ryanzhang I think the dashbuilder-bom should only define artifacts with groupId `org.dashbuilder`. See https://github.com/dashbuilder/dashbuilder/blob/0.2.x/dashbuilder-bom/pom.xml.

It should not introduce (import) any other BOM (e.g. the integration-platform-bom). 

I do agree that the dashbuilder is now an external dependency and such should be moved in that section, I am just not sure if that will resolve the issue you are seeing. Could please share some details?

ryanzhang:
I think It introduced other dependencies via dashbuilder-bom's parent ie,dashbuilder-parent-metadata.
@psiroky here is a real case:
in this build, https://brewweb.devel.redhat.com/buildinfo?buildID=412756
dashbuilder 0.2.0.CR2-redhat-1->mvel 2.2.2.Final-redhat-1
integration-platform-bom CR17-redhat-4 ->mvel 2.2.2.Final-redhat-2
The expectation is let  2.2.2.Final-redhat-2 to win
but the above build finally choose 2.2.2.Final-redhat-1 contribute to the problem I descripted here.

PR #61, opened 2015-04-27
internalizing i18n taglib (so old that cannot be found on maven

porcelli:
central). this will help remove the .tld and .jar from the codebase
of all kie related repos.


manstis:
http://github.com/droolsjbpm/droolsjbpm-build-bootstrap/commit/2f3049bfa

PR #63, opened 2015-05-05
Fix building list of required repositories

yurloc:
* fixed scope of `@prereq_list` variable so that it is visible to
  `prerequisites() sub`
* added detection of dependency cycle with debugging output (which prevents **infinite recursion**)


mrietveld:
Thanks @yurloc! In the future, please feel free to push commits directly for this file.

mrietveld:
(merged on master)

PR #65, opened 2015-05-07
additional dependencies needed for a clean setup of workbenches distros

porcelli:


manstis:
Yikes.. that wasn't the cleanest of merges... must check my "process" for the next one!!

PR #66, opened 2015-05-08
Remove duplicated script + move some scripts to checks/

psiroky:
I believe the moved (Perl) scripts are all checking something, so better place for them is the checks dir. One of the scripts was also duplicated, so I removed it.

@mrietveld I am doing this as PR to be sure I getting the meaning of the scripts right. Please take a quick look and let me know if you agree with these changes.


yurloc:
@psiroky `script/repo-dep-tree.pl` tries to read `./repository-list.txt` so the relative path has to be changed to `../repository-list.txt` when the script is moved.

psiroky:
Thanks @yurloc, this (and few other relative paths) should be fixed now.

mrietveld:
@psiroky Looks good, +1

psiroky:
Merged into master.

PR #68, opened 2015-05-17
maven-source-plugin: use -no-fork goals to avoid duplicated plugin executions

psiroky:
* the 'jar' and 'test-jar' goals invoke execution of whole
   lifecycle phase, which means all the plugins up to that
   phase are executed multiple times during the build.
   That includes for example enforcer and checkstyle plugins.
   See https://maven.apache.org/plugins/maven-source-plugin/plugin-info.html
   for more info.

 * using the -no-fork goals will make sure that only these
   goals are excuted and not the whole lifecycle again.
   The output of the build in target/ dir is the same both
   with and without this change.

@mbiarnes @ge0ffrey please take look just in case I missed something. I believe the change is safe, but its better if someone else reviews the changes before merging.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -353,7 +353,7 @@
             <execution>
               <id>jar</id>
               <goals>
-                <goal>jar</goal>
+                <goal>jar-no-fork</goal>
               </goals>
               <configuration>
                 <archive>
@@ -371,7 +371,7 @@
             <execution>
               <id>test-jar</id>
               <goals>
-                <goal>test-jar</goal>
+                <goal>test-jar-no-fork</goal>
               </goals>
               <configuration>
                 <archive>


mrietveld:
+10 -- especially because phases are no longer repeated!

mbiarnes:
+1 - the change is simple (from test-jar to test-jar-no-fork), the build is not affected and we will save build time

ge0ffrey:
Are those -no-fork goals new? I never heard of them. I wonder why they aren't the default: is there a disadvantage of using them? Or did the developer simply didn't want to take the risk breaking someone's build to improve everyone's build? Let's presume the latter until proven otherwise.

+1 but check if the enforcer plugin does get executed at least once after this change.

I have seen repeating phases before (javadocs, assembly plugin), but I hadn't seen any due to this yet.

psiroky:
Docs mention they were introduced in version `2.1`. I don't know if the behavior of the standard `jar` and `test-jar` goals changed as well.

Both `enforcer` and `checkstyle` plugin are executed once during the build (after the change).

psiroky:
TBH I've never seen these goals before either. I was looking at why some of the plugins are executed multiple times and this seems to be the cause.

mrietveld:
Maven executing phases multiple times is an old "feature-not-a-bug". Getting the link for the issue.. just a sec.

mrietveld:
I ran into this problem for generate-sources targets that I had: 

- [How to Prevent Generate-Sources from executing multiple times](http://blog.peterlynch.ca/2010/05/maven-how-to-prevent-generate-sources.html)

I'll double check that the [kie-remote-jaxb-gen](https://github.com/droolsjbpm/droolsjbpm-integration/tree/master/kie-remote/kie-remote-jaxb-gen) module builds correctly when this is merged.

psiroky:
Thanks for the link @mrietveld! That explains some of the mysteries I am seeing :) It seems the `jboss-parent` is trying to solve the issue mentioned in the article, see https://github.com/jboss/jboss-parent-pom/blob/jboss-parent-16/pom.xml#L662 . It overrides the config for `attach-sources` exec. id, so that only the `-no-fork` goal is called. However, we do have other execution called `jar`, so effective we are generating the -sources jar two times in the build. I will fix also that. Hopefully that's the last issue related to this :)

PR #72, opened 2015-06-05
[DROOLS-788] add kie-platform-bom

psiroky:
* kie-platform-bom should serve as a replacement for
   kie-parent-with-dependencies for cases where we
   need/want all the versions KIE projects are using
   (even the 3rd party ones like sl4j or commons-io).
   The kie-parent-with-dependencies now simply imports
   that BOM and adds the default junit and mockito deps
   (no change there). The plan is to eventually (for
   version 7.x) remove the kie-parent-with-dependencies
   and use only kie-parent and kie-platfrom-bom.

@ge0ffrey @mbiarnes please take a look and let me know what you think.


ge0ffrey:
+1 Makes senses (and is as discussed).

I think the next step after this is removing kie-parent-with-dependencies. Less is more: I think we can no longer justify the existence of kie-parent-with-dependencies. Once it's removed, we can probably just rename kie-parent-metadata back to kie-parent.

psiroky:
One minor issue with removing the `kie-parent-with-dependencies` is that is declares two test-scope dependencies - junit and mockito. So once we remove this parent, we will need to include these deps in all the 20 (or so) top-level POM files.

ge0ffrey:
I can live with that. Not everyone might use mockito anyway.

psiroky:
Yes, this is definitely not a blocker. We just need to keep in mind to upgrade POMs. As we discussed I will plan to remove the `kie-parent-with-dependencies` for 7.0 release. I will create JIRA for that and mention that we need to add the dependencies into the top-level POMs.

mbiarnes:
+1

psiroky:
I'll will update the PR so it is mergeable again, test it locally executing the whole build and commit the changes tomorrow if I don't find any issues.

psiroky:
Finally merged and pushed into master. https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/commit/7eed026fb66f4af54075846a9a1f06b46b3d8643

PR #73, opened 2015-06-08
Selenium dependencies

ederign:


kie-parent-with-dependencies/pom.xml
```diff
@@ -101,6 +101,8 @@
     <!-- version overwrites resteasy in ip-bom - this upgrade has to be done in the jboss-ip-bom> -->
     <version.org.jboss.resteasy>2.3.10.Final</version.org.jboss.resteasy>
 
+    <version.org.seleniumhq.selenium.selenium-java>2.46.0</version.org.seleniumhq.selenium.selenium-java>
```
psiroky: In case there is no ambiguity we should use only the groupId. The artifactId is generally used when multiple dependencies with same groupId have different versions (like for example the org.apache.commons ones).

psiroky:
https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/commit/5dbca3ecb4d607a219f262561e6134fc19804d99

PR #76, opened 2015-07-01
Add Issue Keeper dependency

livthomas:
Issue Keeper will allow us to connect tests with BZs or JIRAs using simple annotations. The test with such annotation is not run unless the related issue is resolved.

This is mostly used in jBPM engine tests which are being moved from QE test suite to jBPM repository and depend on the states of the related BZs.


mswiderski:
+1

PR #77, opened 2015-07-10
Enable code coverage analysis with JaCoCo

yurloc:
Adding JaCoCo configuration to `kie-parent-metadata` makes code coverage analysis ready to use in any droolsjbpm project. I find this potentially useful for both QA engineers and developers who want to quickly find code pieces lacking test coverage.

I have introduced a simple rule that requires 90% line coverage. The plugin wil print a warning if the requirement is not met. This can be changed or the check execution can be removed completely. More interesting is the HTML report of code coverage that is generated after running

```shell
mvn clean verify -Djacoco
```

The reports can be found in `target/site/jacoco/index.html` in any module that executed some tests.

The test execution analysis and subsequent report generation may add some 10-50% to build time (very roughly) so I put JaCoCo plugin into a profile that is not active by default. It can be activated with `-Djacoco`.

Surefire upgraded so that JaCoCo agent path passed in `argLine` can be
combined with memory options for test execution ([SUREFIRE-1047](https://issues.apache.org/jira/browse/SUREFIRE-1047) fixed in 2.17).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -164,6 +164,9 @@
     <latestReleasedVersionFromThisBranch>notYetReleased</latestReleasedVersionFromThisBranch>
     <!-- Version of the KIE Workbench application. Shown for example in About dialog. Will be usually overriden by productisation. -->
     <version.org.kie.workbench.app>${version.org.kie}</version.org.kie.workbench.app>
+
+    <!-- Set default argLine value to empty to prevent surefire failure when jacoco is disabled -->
+    <surefire.argLine></surefire.argLine>
   </properties>
 
   <repositories>
@@ -417,7 +420,7 @@
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-surefire-plugin</artifactId>
-          <version>2.15</version>
+          <version>2.18.1</version>
           <configuration>
             <includes>
               <include>**/*Test.java</include>
@@ -425,7 +428,8 @@
             <excludes>
               <exclude>**/*IntegrationTest.java</exclude>
             </excludes>
-            <argLine>-Xmx1024m -XX:MaxPermSize=128m -Dfile.encoding=UTF-8</argLine>
+            <!-- Append surefire.argLine property populated by JaCoCo plugin -->
+            <argLine>-Xmx1024m -XX:MaxPermSize=128m -Dfile.encoding=UTF-8 @{surefire.argLine}</argLine>
             <systemProperties>
               <property>
                 <name>apple.awt.UIElement</name>
@@ -442,7 +446,7 @@
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-failsafe-plugin</artifactId>
-          <version>2.15</version>
+          <version>2.18.1</version>
           <executions>
             <execution>
               <goals>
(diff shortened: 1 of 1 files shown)


mrietveld:
@psiroky Could you review this, given that you have more knowledge of the build?

psiroky:
@mrietveld yes, I will take a look at this today.

psiroky:
I haven't had time to look at this yet, but one thing that come to my mind: could we use something like `-Dtest-covereage` or maybe just `-Dcoverage`? So that in case we would switch to something different than JaCoCo, we could still use this general name. WDYT @yurloc?

pom.xml
```diff
@@ -475,6 +479,48 @@
...
+                <!-- Do not fail the build if rules are not met -->
+                <haltOnFailure>false</haltOnFailure>
+                <rules>
+                  <rule>
+                    <element>BUNDLE</element>
+                    <limits>
+                      <limit>
+                        <counter>LINE</counter>
+                        <value>COVEREDRATIO</value>
+                        <minimum>0.9</minimum>
```
psiroky: I think this is too strict. But since we are just reporting warnings, it probably does matter that much.

pom.xml
```diff
@@ -181,6 +181,9 @@
     <!-- Remove hibermate-jpamodelgen prop when we move to jboss-integration-platform-bom 6.0.0.CR27+ -->
     <!-- When we move to hibernate 4.3+: version.org.hibernate.jpamodelgen should be removed and version.org.hibernate used instead  -->
     <version.org.hibernate.jpamodelgen>1.3.0.Final</version.org.hibernate.jpamodelgen>
+
+    <!-- Set default argLine value to empty to prevent surefire failure when jacoco is disabled -->
+    <argLine></argLine>
```
psiroky: Maybe we could use something like `surefire.argLine`? Or does it apply to other plugins as well?

yurloc:
@psiroky Ad `-Djacoco` vs. `-Dcoverage`: From my point of view using "jacoco" property is easier to remember. Maybe the profile won't be necessary any more if we switch the coverage tool. Or we might want to use different coverage tools together, each enabled with its own profile.

But I see your point. Once the name starts to be used in different places, changing it may be painful. I like `-Dcoverage`. I will update that if you confirm it.

pom.xml
yurloc: Yes, it is strict. We could certainly have a long discussion about the rules... I came across a statement that I liked:

> There is no reason to have less than 95% of lines covered by tests when doing TDD.

Or something like that. I think it's true. Rhetorical question: Do we really have code that is never executed during testing? Are we OK about that?

On the other hand, who is doing TDD here? Additionally, there is some code where the added value of tests is questionable (e.g. GWT client code, maybe some other...).

So, this rule is basically a demo that will print a warning with the actual line coverage (if it's below 90%) so anyone who's running a build can either investigate the report and maybe make some improvement or just ignore it. I think there is no "one size fits all" so if we want the `check` goal to be taken seriously, or even have it fail the build, rules will have to be tailored in different ways for different projects.

manstis: I agree with Jiri.... but woe be the day we fail builds for lack of test
coverage ;-)

Rikkola: I'll confess, I'm doing TDD most of the time and not even meaning Toni Driven Development, but the real thing ;)

Like Jiri mentioned, 90% will be a problem with the MVP view implementations we have. We only test the presenter since, the view is GWT and unit testing it is hard, but we also keep the views as simple as possible ( = just setting values or styles ). The view part is still Java code, not sure if it will be possible to ignore *ViewImpl.java files when counting the %?

ge0ffrey:
Why add "-Djacoco" (and a special profile to make it work), why not simply call the goal directly "mvn jacoco:jacoco" (or "mvn clean install jacoco:jacoco")?

psiroky:
Yes, that should work as well. We are already doing that with findbugs.

psiroky:
@yurloc what do you think about using just `jacoco:jacoco` like @ge0ffrey suggested? It uses the specific names, but it also means we don't have to use additional profile, which is good.

yurloc:
> Why add "-Djacoco" (and a special profile to make it work), why not simply call the goal directly "mvn jacoco:jacoco" (or "mvn clean install jacoco:jacoco")?

@ge0ffrey There is no "jacoco" goal in jacoco plugin. A meaningful usage of the plugin involves 2 stages:

1. `prepare-agent` goal runs before test execution. As a result runtime agent is attached to surefire and code coverage analysis is produced after tests are executed

2. At least on of the following (after test phase):
  - `check` verifies that code coverage meets defined criteria
  - `report` produces a graphical (HTML) report of the code coverage

It is possible to get the analysis with one Maven execution (`mvn clean test`) and that generate report in another execution (`mvn jacoco:report`). On the other hand, `mvn jacoco:check` fails due to missing `rules` configuration (it seems that POM configuration is ignored when invoking the goal like this). All in all it seem the only possible way to perform both `check` and `report` is to run `mvn clean verify`.

Regarding the new profile -- its purpose is to make code coverage analysis and report generation optional so that it doesn't slow regular builds.

ge0ffrey:
That's a pity that the maven-jacoco-plugin doesn't have such a facade-pattern goal (IIRC the corbertura plugin does - not sure about the emma plugin). But in that case, I agree adding the profile is the lesser evil.

yurloc:
Made couple of changes:
* renamed the property used to carry over jacoco agent path to `surefire.argLine` (to prevent possible collisions with other plugins?)
* changed activation property name to `code-coverage`

psiroky:
Merged into master https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/commit/2954e8216939b483f850bad84905a20d90a1b231.

Cherry-picked into 6.3.x https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/commit/9b9776b268f089a8669b3bd92c49ebc4b0ffd11c

PR #82, opened 2015-07-22
Update errai-cdi10 version to 3.0.5.Final to match support version

ryanzhang:
The latest Errai support version on 3.0.x is 3.0.5.Final-redhat-1. So let's update to 3.0.5.Final.
I hope that's ok.


psiroky:
Ryan, I think we should upgrade directly to 3.0.6.Final, which was released just yesterday. WDYT?

ryanzhang:
So let's say 6.1.2.CR3 or 6.1.3(future release) will also update to errai 3.0.6.Final , right?
If that's the case, we should do that.

psiroky:
Yes, the next build of 6.2.x branch will use Errai 3.0.6.Final.

ryanzhang:
Petr, The errai_cdi10 has been update to 3.0.6.Final.

PR #83, opened 2015-07-22
Move release related sections into own README file

psiroky:
* the release related sections are not important for
   most of the contributors, so there is no need
   to include them in default README (it is big enough
   even without them).


psiroky:
Merged into master https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/commit/ae34c378ae4e4e1ad88a2aac03278f5864aa147a and cherry-picked into 6.3.x

PR #85, opened 2015-08-27
Update README.md

lmolinaro:
Adding resolution in case of error while building project with maven when someone are using jdk8.

### DIFF
--- a/README.md
+++ b/README.md
@@ -401,7 +401,7 @@ Running the build
         
         b.  set global variable JAVA_HOME=(pointing to jdk7)
         
-        c.  *optional if you get OutOfMemoryError* set java VM size MAVEN_OPTS="-d64 -Xms256m -Xmx1024m -XX:MaxPermSize=512m"
+        c.  *optional if you get OutOfMemoryError* set java VM size MAVEN_OPTS="-d64 -Xms512m -Xmx1024m -XX:MaxPermSize=1024m"
             c.1. tested on a 64 bit machine, from here -d64 parameter.
         
         c.  run maven command to build again.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -389,6 +390,23 @@ Running the build
     and you are behind a non-transparent proxy server,
     [configure your proxy server in Maven](http://maven.apache.org/settings.html#Proxies).
 
+ Note: if you get the following error: 
+ 
+    ```shell
+    [ERROR] Failed to execute goal org.codehaus.mojo:jaxb2-maven-plugin:1.5:schemagen (generate-xsd) on project kie-remote-jaxb-gen: Failed to generate schema: InvocationTargetException: Class not foundcom/sun/tools/apt/Main.class -> [Help 1]
+    ```
```
psiroky: This should have been fixed by https://github.com/droolsjbpm/droolsjbpm-integration/commit/73d1130f3921df16ed8ab2ee98baff87f5868a75. I just tried locally and everything works for me with Oracle JDK 1.8.0_51. If the problem persist, please let us know more details. We need to fix this issue, instead of providing workarounds.

lmolinaro: oooh.. I have needed 6.2.x branch and that resolution is not present there. Maybe, is better to add that resolution also in that branch?

psiroky: The 6.2.x branch is no longer supported. There will be no more releases from that branch, so even if we cherry-picked the change, there would be no new release with that fix. If you are building from sources (which you do it seems), you can easily cherry-pick that commit into your own branch.

lmolinaro:
thanks, for all these tips

PR #87, opened 2015-09-13
Remove latestReleasedVersion property

psiroky:
* this property is not used by productisation anymore

@ryanzhang can you please confirm that you don't need this property?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -163,8 +163,6 @@
     <!-- enabling the next line will do that the build will not break when an illegal transitive dependency is found,
          the default is: the next line commented -->
     <!-- <illegaltransitivereportonly>true</illegaltransitivereportonly> -->
-    <!-- property for productisation to know the last released version -->
-    <latestReleasedVersionFromThisBranch>notYetReleased</latestReleasedVersionFromThisBranch>
     <!-- Version of the KIE Workbench application. Shown for example in About dialog. Will be usually overridden by productisation. -->
     <version.org.kie.workbench.app>${version.org.kie}</version.org.kie.workbench.app>
   </properties>


ryanzhang:
Can we put the latestReleasedversion to handoff doc[1] if removed in pom.xml? 
Actually this information is required for definining the product artifact version. But agreed that it can be removed if we define in in the handoff doc. 

What do  you think?
[1]: https://github.com/jboss-integration/kie-release-reports/tree/master/reports/tags

ge0ffrey:
If productization doesn't use this property, then how do the figure out what the correct latest community release version is?
  https://bugzilla.redhat.com/show_bug.cgi?id=1168653

psiroky:
I am OK with having the property in the pom.xml, it should be easier to maintain than in the hand-off docs.

The reason I thought it is not used is because on 6.3.x branch (https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/blob/6.3.x/pom.xml#L165) , the value is still `notYetReleased`, even though we had more than few releases from that branch (both community and product) and I haven't heard any complaints about that.

ryanzhang:
Yes, ideally we need to explicit define the latest version somewhere where productization can know the upstream release version. (pom.xml or handoff, it really denpends on your convinence).
For the previous few releases, I am tracking the community release. So I would make it correct even it is not there. However if someone else from productization do the prod  build, he might not able to tell the correct version if it's not there.
So the best case is do the declare explicitly. Just let me know which way is the best to your maintainence.

psiroky:
We can maintain the property in pom.xml, that is fine. But please do scream if you see it is not correctly set. Otherwise we will likely get to the same situation we are in now. I will see if we can do the property updates automatically to avoid manuall effort which brings errors.

psiroky:
I am closing the PR since the property is still needed (see above discussion).

ryanzhang:
OK, will do.
Thanks!

PR #90, opened 2015-09-28
Use mockito-core instead of mockito-all (which is an uberjar)

psiroky:
@mbiarnes, @ge0ffrey do you remember if there was a reason to use `mockito-all`? It is an uberjar and those should not be used at all. `mockito-core` properly specifies the other dependencies, so I believe this change should be transparent.


ge0ffrey:
I didn't even know we were using an uberjar. We definitely shouldn't do that.
 +1 to replace it with mockito-core if all (test) code still compiles.

mbiarnes:
+1 - if it compiles

PR #97, opened 2015-11-05
[6.3.x] Upgrade Errai to 3.2.2-SNAPSHOT

psiroky:


manstis:
+1 to merge.

csadilek:
I am wondering if we really should do this for 6.3.x? There are currently no plans for Errai 3.2.2.Final and moving to 3.2.2-SNAPSHOT would require a 3.2.2.Final release in November. Is there anything you need from Errai?

psiroky:
I am completely fine with upgrading to 3.2.1.Final. We definitely can not depend on 3.2.1-SNAPSHOT.

csadilek:
Yeah, I think that's better in this case as we're not expecting any needs for Errai for the 6.3. CR/Final releases.

psiroky:
OK. I will create new PR with upgrade to 3.2.1.Final. (I can't amend this one as the branch was created in blessed repository and it does not allow force pushes.)

PR #99, opened 2015-11-05
update-version-all: support both community and prod builds

psiroky:
@mbiarnes please take a look and review.

### DIFF
--- a/script/release/update-version-all-community-settings.xml
+++ b/script/release/update-version-all-community-settings.xml
@@ -11,15 +11,18 @@
   <proxies/>
   <profiles>
     <profile>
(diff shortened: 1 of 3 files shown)


mbiarnes:
+1
think about that we have to change all Jenkins CI jobs which use update-version-all.sh to use the right command: 
../droolsjbpm-build-bootstrao/scrpt/release/update-version-all.sh <release versions) community (or prod)

psiroky:
Yes, we should gradually update all the jobs. But for now it will just work as it is, because the changes are backwards compatible.

psiroky:
Merged into master and 6.3.x.

PR #100, opened 2015-11-06
Remove dependencies already prenset in IP BOM

psiroky:
* all those deps are already in ip-bom:6.0.0.CR29
   with exactly the same versions

@mbiarnes please take a look.


mbiarnes:
yepp - these deps are in jboss-ip-bom 6.0.0.CR30 too

psiroky:
Merged into master and 6.3.x

PR #101, opened 2015-11-06
git-clone-others.sh: support URLs not ending with .git

psiroky:
* when cloning from github, one can use both github.com/org/repo
   and github.com/org/repo.git. This change add supports for the
   former (without .git suffix) URL

@mbiarnes this is a trivial change, but please double check anyway.


mbiarnes:
+1 - this works

psiroky:
Merged into master and 6.3.x.

PR #105, opened 2015-11-17
changed prod to product in update-version-all.sh for matching right v…

mbiarnes:
…alue of Jenkins script variable

For matching values of variables in Jenkins CI scripts for releasing it is desired that this variable is product instead of prod.


psiroky:
Replace by https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/107

PR #107, opened 2015-11-24
changed prod to productized in update-version-all.sh for matching rig…

mbiarnes:
…ht value of Jenkins script variable


script/release/updateTest.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
```
psiroky: It seems this file was committed by mistake?

mbiarnes:
it is easier and faster to close thsi PR and raise a new one

PR #110, opened 2015-11-27
maven-bundle-plugin: disable OBR

psiroky:
* updates to local OBR (OSGi Bundle Repository) are not
   necessary and they only prolong the build if the repository
   (by default ~/.m2/repository/repository.xml) is large

@mariofusco please take a look as this is related to OSGi


ge0ffrey:
Do drools and optaplanner still build with an empty maven repository after this change?

psiroky:
I haven't tried that as it takes ages (but just started local build to make sure). I tried building after removing ~/.m2/repository/repository.xml and that worked fine.

psiroky:
Confirmed. Everything builds fine, even when starting with empty localy maven repo.

ge0ffrey:
+1 (for the record)

PR #115, opened 2015-12-05
Update copyright holder name

richardfontana:
JBoss Inc. has not actually existed since 2006 :)
Here I suggest updating the copyright notices in this repository to refer to 'Red Hat, Inc. and/or its affiliates', if only because this is providing guidance to droolsjbpm projects on how to handle copyright notices.

Note that a Red Hat-specific copyright notice is not necessarily the best choice. Another possibility which you may wish to consider would be to refer generally to contributors to $projectname, e.g. 'Copyright $year Drools Authors'.

### DIFF
--- a/README.md
+++ b/README.md
@@ -768,7 +768,7 @@ Configuring Eclipse
 
         ```
         /*
-         * Copyright 2015 JBoss Inc
+         * Copyright 2015 Red Hat, Inc. and/or its affiliates.
          *
          * Licensed under the Apache License, Version 2.0 (the "License");
          * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


psiroky:
Thanks Richard! I merged the changes as you suggested. It is probably easier like that since we have multiple projects (Drools, jBPM, OptaPlanner) and we would need to use different headers for those if we went with `$project Authors`.

psiroky:
Now we only need to fix those hundreds of files using the old header :) Fortunately, that should be easily scriptable.

richardfontana:
I was thinking that at least going forward the updated header could be used for new files, but of course if you want to fix all the old ones that is fine. :)

ge0ffrey:
We've send a mail to all developers to ask them to update their template.

Michael Biarnes will change all headers on master on the droolsjbpm repo's next week (to give the developers some time to update their templates first and push their local changes to master).
Currently, we don't intend to fix old branches, probably not even 6.3.x.

@richardfontana Thanks for this PR :) If you see anything else that needs to change for legal reasons in our code or community websites (drools.org, optaplanner.org, jbpm.org), let us know (preferably through a PR again).

ge0ffrey:
@richardfontana Should Java files that used in a maven archetype (which generate a hello world project for our users) also contain the license header? What about java files in examples?

If they don't contain it, then there's no "without warranty" clause.
If they do they contain it, then users start asking if they can change the license in their project that was generated from the maven archetype (or copy pasted from the examples): https://issues.jboss.org/browse/DROOLS-599 (and of course they need to change the license from their boss).

richardfontana:
@ge0ffrey I would suggest not including the license header in those cases.

ge0ffrey:
Thanks!

PR #120, opened 2015-12-16
[BZ-1199965] upgrade to ecj 4.4.2 in order to support Java 8 syntax in drl files

mariofusco:


psiroky:
We should remove the ECJ from kie-platfrom-bom as well (it will be brought by ip-bom): https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/blob/master/kie-platform-bom/pom.xml#L421

mbiarnes:
+1 - tested locally

mariofusco:
@psiroky done

PR #122, opened 2016-01-05
DROOLS-1019 add kmodule.xsd publishes to the release procedure

ge0ffrey:


ge0ffrey:
See also https://issues.jboss.org/browse/DROOLS-1019 for motivation why we need those XSD's there.

@mbiarnes @psiroky Let me know if the new step during the release procedure isn't clear

psiroky:
Thanks @ge0ffrey, the steps are clear. I think we should automate this, otherwise there will always be chance that we forget....

ge0ffrey:
@psiroky Can you merge?

psiroky:
Merged. @mbiarnes it would be great if you could take a look at automating this.

PR #124, opened 2016-01-06
Updating the oldbranches.pl script

mrietveld:
Minor updates/fixes to make the half-yearly or yearly checks easier.

### DIFF
--- a/script/checks/oldbranches.pl
+++ b/script/checks/oldbranches.pl
@@ -2,10 +2,34 @@
 
 use DateTime;
 use Cwd qw(abs_path);
+use Getopt::Std;
 
 use POSIX;
 use strict;
 use warnings;
+
+my %opts;
+getopts( 'vhr:', \%opts );
+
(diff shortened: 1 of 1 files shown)


mrietveld:
@psiroky Would you mind looking at this?

psiroky:
yay, Perl stuff :) I will take a look (and test locally).

mrietveld:
O, hold on, I need to add a die/fail message for the -r option.

mrietveld:
Okay, fixed! 
@psiroky Please review!

psiroky:
+1. Merged into master.

One thing I noticed is that the script excludes some branches which probably should not be excluded, like `uf_0.5.x` or `lienzo-core-2.0.x`. I think we could check if the branch starts with number and ends with `x` (this the pattern that we follow for long-running branches). But this really is just minor thing. Thanks @mrietveld for the changes, this script is very useful!

PR #126, opened 2016-01-08
Upgrade maven-filtering to 1.3 for resources-plugin, to fix MSHARED-325

psiroky:


ge0ffrey:
+1, we 're at the beginning of a development cycle, so all maven plugins should be updated to their latest version (except those for which there's a good reason not to do so, such as JDK 6 compatibility).

psiroky:
The upgrade of jboss-parent (transitively via ip-parent) was done recently. I believe most plugins we use are at latest (or latest but one) versions. I would like to keep plugin versions aligned to jboss-parent as much as possible as it that is already used by WildFly (so there is much lower chance something will break) and it should also make lives easier for prod. guys as they only need to take care of one set of plugins in their prod. environments. WDYT?

PR #129, opened 2016-01-19
Upgrade Cargo plugin version to 1.4.18

sutaakar:
Cargo is used just for integration tests, raising of version allows to create WebSphere standalone profiles.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -442,7 +442,7 @@
         <plugin>
           <groupId>org.codehaus.cargo</groupId>
           <artifactId>cargo-maven2-plugin</artifactId>
-          <version>1.4.14</version>
+          <version>1.4.18</version>
         </plugin>
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>


mswiderski:
+1 as it makes the WebSphere tests way easier to run (same as other containers)

psiroky:
@ryanzhang could you please take a look at this and let us know if the plugin upgrade can cause some issues in the product 6.2.x build? (e.g. some additional work needed to import the new version to the product build env).

ryanzhang:
Yes, we will need to import the new plugins for this changes.
I have no objection here.
Could you let me know which project is actually using the Cargo plugin? So I can find all dependant things to add .

psiroky:
@ryanzhang thanks for the quick reply, I wll merge this then. The Cargo is used in droolsjbpm-integration and kie-wb-distributions (e.g. kie-wb-smoke-tests) for testing on different containers.

ryanzhang:
Thanks for your information, Petr.

PR #131, opened 2016-01-27
Enforce Java 8+ and remove PermGen settings

psiroky:
@csadilek please merge this together with your Errai 4 upgrade.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -127,16 +127,15 @@
          See https://issues.apache.org/jira/browse/FELIX-4882 for more details.
          Version 3.x requires Java 7 (not directly, but one of transitive deps does). -->
     <version.bundle.plugin>2.5.3</version.bundle.plugin>
-    <!-- Override6.1.1 is the last version supporting Java 6. We need to be able to run certain modules
-         (e.g. jbpm-test-coverage) with Java 6. -->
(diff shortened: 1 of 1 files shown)


ge0ffrey:
Does anyone else feel like celebrating when they see this? :)

psiroky:
And the Jenkins the is not happy at all :) I will need to create new PR jobs for master and use Java 8 there.

psiroky:
jenkins retest this

psiroky:
@csadilek I've updated the PR, it now also removes the animal-sniffer config which is not needed anymore (should be removed from uf and dashbuilder as well)

psiroky:
@csadilek please don't merge this yet. I just found out that drools won't compile with these changes (looks like changing the source+target levels has something to do with this). You can still push you Errai 4 upgrade, but we should not increase the default source + target levels yet. I will let you know once I manage to fix it.

psiroky:
@csadilek I've fixed all the build-related issues I know of, so please go ahead and merge when doing the upgrade.

csadilek:
+1 (on the way)

csadilek:
master: 955f539bc5ee73069c511dc851994efb06c0381e

PR #134, opened 2016-01-27
Upgrade FindBugs plugin to 3.0.1 to support JDK 8

psiroky:


psiroky:
I am merging this directly without +1 as it is trivial build related change and we need to fix the CI builds (otherwise lots of false negative emails will be sent).

ge0ffrey:
ok to merge with +1, but would you mind mentioning both the from and the to version in future commit messages? That gives an indication if it's a hotfix upgrade or something more.

psiroky:
@ge0ffrey I am usually doing that, just forgot this time for some reason. Will make sure to do that next time. The previous version was 2.5.5. We need version 3.x which is the first one supporting Java 8.

PR #136, opened 2016-01-28
[DO NOT MERGE] testing Jenkins CI

sometestuser1234:


psiroky:
Please don't close this. I created it with account that is not in droolsjbpm org. unit to see how Jenkins copes with that.

psiroky:
ok to test

PR #138, opened 2016-01-28
Update Errai to 4.0.0-SNAPSHOT, GWT 2.8.0-beta1, and Java 8.

mbarkley:


csadilek:
+1

csadilek:
master: 17bc176fdb99209471ced2a609f9f3d0c5836158

pom.xml
```diff
@@ -102,7 +103,8 @@
     <version.org.simpleframework>6.0.1</version.org.simpleframework>
     <!-- this version will overwrite jboss-ip-bom version 2.5.1. The version in jboss-ip-bom couldn't be upgraded as it will break ModeShape Web Explorer application which uses GWT 2.5.1 -->
     <!-- please look at https://github.com/fredsa/gwt-dnd/wiki/GettingStarted -->
-    <version.com.google.gwt>2.7.0</version.com.google.gwt>
+    <version.com.google.gwt>2.8.0-beta1</version.com.google.gwt>
+    <version.com.google.guava>20.0-SNAPSHOT</version.com.google.guava>
```
psiroky: Do we really need Guava 20? I think we should avoid using 3rd party SNAPSHOTs as that may bring unpleasant surprises.

mbarkley: I also really dislike this, but unfortunately there is no released version of guava that is compatible with GWT 2.8. Using guava 19 (the most recent version) causes build errors. There is literally no stable version of guava we can use.

psiroky: Seems like we are really using bleeding edge stuff :) I guess we won't be doing any releases from master for some time, so should be acceptable for now.

PR #139, opened 2016-01-28
Use released version of Guava: 19.0.0.jbossorg-1

psiroky:
* this is a custom Guava build taken from current master.
   It is called 19* because it really is just 19 + bunch
   of fixes. Calling it 20* would be more confusing.


csadilek:
+1

csadilek:
master: 4c7b767ee8a9a89dc6f2a00465caf182829e8dfd

PR #140, opened 2016-01-29
Add instructions about inotify on Linux

manstis:
Instructions learned following the move to GWT2.8 etc

### DIFF
--- a/README.md
+++ b/README.md
@@ -1149,6 +1149,26 @@ Extra IntelliJ plugins
 
         * Run that configuration.
 
+Linux inotify
+-------------
+
+We have encountered some issues with different Linux distribution's *inotify* when running any of the Workbench's in Super Dev Mode from within InteliJ.
+
+Error message `User limit of inotify instances reached or too many open files` has been observed with both Ubuntu and Fedora. Should you encounter this issue you will need to reconfigure your *inotify* settings.
+
+Add the following to `/etc/sysctl.conf` and then run `sudo sysctl -p`:
+
+`fs.inotify.max_user_watches = 524288`
+
+`fs.inotify.max_user_instances = 524288`
+
+You may also need to add the following lines (replacing `user-id` with your User Id) to `/etc/security/limits.conf`:
+
+`user-id soft nofile 4096`
+
+`user-id hard nofile 10240`
+
+
 Team communication
 ==================
 


README.md
```diff
@@ -1149,6 +1149,20 @@ Extra IntelliJ plugins
 
         * Run that configuration.
 
+Linux inotify
+-------------
+
+We have encountered some issues with Linux's *inotify* when running any of the Workbench's in Super Dev Mode from within InteliJ.
+
+Error message `User limit of inotify instances reached or too many open files` has been observed with Ubuntu 15.04 (but not with Fedora 22). Should you encounter this issue you will need to reconfigure your *inotify* settings.
```
csadilek: We've actually been able to reproduce this in Fedora as well.

README.md
```diff
@@ -1149,6 +1149,20 @@ Extra IntelliJ plugins
...
+
+Add the following to `/etc/sysctl.conf` and then run `sudo sysctl -p`:
+
+`fs.inotify.max_user_watches = 524288`
+
+`fs.inotify.max_user_instances = 524288`
+
+
```
csadilek: There's a related issue where the IntelliJ GWT tooling is opening too many files (as opposed to inotify instances). This can be fixed like this:

user soft nofile 4096
user hard nofile 10240

csadilek:
@manstis just added some comments. I can make the changes as well, if you prefer.

manstis:
@csadilek Next time; I'll check my RH email before updating the document ;)

csadilek:
+1, thanks.

csadilek:
master: 0097c032228dcaa019d96beb337010ad49429d17

ge0ffrey:
Is this the same error as IntelliJ warns about on the bottom left?
If it is, IIRC it also notes that a specific version of a specific library will fix this sometime in the future. So - if possible - we might want to mention that library name/version in this instruction so we know when it's no longer applies.

manstis:
Bottom left? Message? You'll need to be a bit more specific. If you mean
idbus; then no this is completely different.

ge0ffrey:
Yes, I was thinking it might be idbus.

Anyway, thanks for writing this up :) I am also running into when starting drools-wb (but not with optaplanner-wb, even though they have the same mvn and gwt dependencies - strange)

PR #142, opened 2016-02-11
Move optaplanner-wb to just below drools-wb

ge0ffrey:
because jbpm human tasks might some day use optaplanner which uses drools


psiroky:
This should be safe to do, so +1.

csadilek:
master: 4c47f3bdfe2c417382df36990921d2fa8b9b7220

PR #148, opened 2016-02-28
[6.4.x] Upgrade Errai from 3.2.2-SNAPSHOT to 3.2.3-SNAPSHOT

psiroky:
@csadilek, @mbarkley maybe we should just upgrade to 3.2.2.Final? In any case, we should not depend on 3.2.2-SNAPSHOT anymore.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -89,7 +89,7 @@
     <version.org.jboss.xnio>3.2.0.Final</version.org.jboss.xnio>
     <version.org.apache.xmlbeans>2.3.0</version.org.apache.xmlbeans>
     <version.org.drools.guvnor-api>5.6.0.Final</version.org.drools.guvnor-api>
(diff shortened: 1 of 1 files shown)


mbarkley:
@psiroky, @csadilek We will need to release Errai 3.2.3.Final and upgrade to that (for the Errai UI i18n properties file support).

psiroky:
@mbarkley thanks for the info! So in the mean time we could swtich to 3.2.3-SNASPHOT, to make sure we test with the latest bits (unless you are planning the release in day or two)?

mbarkley:
@psiroky Yeah that would be a good idea. I can open a new PR when we release 3.2.3.Final.

csadilek:
+1

csadilek:
6.4.x: 3b8e7d5ea82a953cb64a88c6e0d72920041234d0

PR #157, opened 2016-03-16
BZ-1310647 update groupId of org.beanshell to org.apache-extras.beans…

mbiarnes:
…hell

### DIFF
--- a/kie-platform-bom/pom.xml
+++ b/kie-platform-bom/pom.xml
@@ -305,6 +305,13 @@
         <version>${version.link.bek.tools.issue-keeper-junit}</version>
       </dependency>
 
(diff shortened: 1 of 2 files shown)


ge0ffrey:
IIRC beanshell is one of those dependencies that we should get rid of due to security reasons (the last release was over a decade ago). Which poms are still using it?

        mvn-all.sh dependency:tree | grep -n10 beanshell

psiroky:
It seems it has been (partially?) resurrected, see https://github.com/beanshell/beanshell. Although I agree we should get rid of it if possible.

psiroky:
Closing, replace by https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/166

PR #160, opened 2016-03-22
Upgrade Cargo plugin version to 1.4.19

sutaakar:
New version of Cargo plugin is released: https://codehaus-cargo.atlassian.net/wiki/pages/viewpage.action?pageId=19103755

Among other things it brings improved handling in case of container startup or deployment failure, see: https://codehaus-cargo.atlassian.net/browse/CARGO-1375


psiroky:
Nice, +1! Thanks @sutaakar!

PR #163, opened 2016-03-29
JBPM-4979 - Generated PDF is unreadable

tsurdilo:
Adding yahp used by jBPM Designer to create pdf from html

### DIFF
--- a/jbpm-bom/pom.xml
+++ b/jbpm-bom/pom.xml
@@ -865,6 +865,46 @@
         <version>${version.org.jbpm}</version>
         <classifier>sources</classifier>
       </dependency>
+      <dependency>
+        <groupId>org.allcolor</groupId>
+        <artifactId>org.allcolor.yahp</artifactId>
+        <version>${version.org.allcolor.yahp}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.allcolor</groupId>
+        <artifactId>org.allcolor.yahp.internal</artifactId>
+        <version>${version.org.allcolor.yahp}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.allcolor</groupId>
+        <artifactId>org.allcolor.yahp.renderer</artifactId>
+        <version>${version.org.allcolor.yahp}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.allcolor</groupId>
+        <artifactId>org.allcolor.yahp.itext</artifactId>
+        <version>${version.org.allcolor.yahp}</version>
+      </dependency>
+      <dependency>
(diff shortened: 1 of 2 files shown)


psiroky:
-1.

 - these dependencies were manually uploaded into repository.jboss.org. The POMs do not contain proper dependencies section! See e.g. https://repository.jboss.org/nexus/content/groups/public/org/allcolor/org.allcolor.yahp.itext/1.5/org.allcolor.yahp.itext-1.5.pom. I don't think this is acceptable.

- I was not able to find sources. I googled https://github.com/allcolor/YaHP-Converter, but the last tag there is 1.3 and the last commit is 4 years old. Are there sources for the 1.5 version?

- that fact that the library is not in Maven Central is alarming. It suggests the library is not being properly maintained, nor widely used. (but I may be wrong).

ge0ffrey:
Are there any versions that are compatible with our dependency requirements?
  https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/blob/master/README.md#requirements-for-dependencies

tsurdilo:
@psiroky I built the jars. They do not have any dependencies as are self-contained individually so is not needed. The version 1.5 is because jboss thirdparty repo uploads were so flaky over the weekends it did not upload some of my jars but then there is no way to add to existing and I had to up a version in order to have it there.

tsurdilo:
@ge0ffrey not that i know of

tsurdilo:
@psiroky the library works and it actually works great for what is needed for the jbpm designer process docs. there would be an alternative which is to pull in itext and flying saucer libs and write a whole lib like this ourselves but that would take a lot of time.

krisv:
@psiroky I think your concern on sources and dependencies etc. is a valid one.  It seems this 1.5 version is actually the same as the 1.3 community version, but mavenized.  What we could do for example is this: we set the version to 1.3-jboss-1 and include a readme in the jar that describes the exact changes).  And add correct dependencies information.  Wdyt?

tsurdilo:
I will try to mavenize thee libs used and do another PR

psiroky:
Assuming the upstream version is 1.3 and we did not do any patches on top of that, we should name our release 1.3.jbossorg-0 (see https://developer.jboss.org/wiki/MavenDeployingAThirdpartyRelease).

My other concern is that the 1.3 version is 4 years old, which is quite a long time. Do we know if the library is still being maintained? What happens if there is security fix needed, will the upstream developer(s) provide it? Or will have to do that ourselves?

There are very good reasons to be very cautious when adding new deps. We already depend on few (half) dead projects, so let's make sure we do not add new one in the list.

PR #164, opened 2016-03-29
JBPM-4979 - Generated PDF is unreadable

tsurdilo:
Adding yahp used by jBPM Designer to create pdf from html


tsurdilo:
I will try to mavenize thee libs used and do another PR

PR #166, opened 2016-04-01
Upgrade to ip-bom 7.0.0.CR1

psiroky:

PR #168, opened 2016-04-04
- DROOLS-972: adding deltaspike core and ctrlcdi-weld dependencies

Salaboy:


Salaboy:
I'm not sure about the definition of deltaspike dependency in our parent pom. Mostly because it is also defined in the jboss-integration parent pom, but it doesn't seem that we can use those variables in our projects.

psiroky:
You can reuse the property coming from https://github.com/jboss-integration/jboss-integration-platform-bom/blob/master/pom.xml#L191

Salaboy:
For some reason I couldn't, but in theory if I just remove the variable definition should it work?

psiroky:
Oh, my bad. This will be possible once we merge #166. Then you can simply remove the property....

mariofusco:
Merged by https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/commit/ff984b29f

PR #173, opened 2016-04-08
Upgrade Weld from 1.1.31.Final to 2.3.3.Final and Weld API from 1.1.Final to 2.3.SP1

psiroky:
* this is a temporary solution until we upgrade to ip-bom which contains the same
   upgrade

### DIFF
--- a/kie-platform-bom/pom.xml
+++ b/kie-platform-bom/pom.xml
@@ -439,6 +439,12 @@
         </exclusions>
       </dependency>
 
+      <dependency>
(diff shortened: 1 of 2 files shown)


psiroky:
Hmm, this breaks CDI testing in drools-compiler, need to take a look at that.

psiroky:
We need to merge all the Weld related PRs at the same time, otherwise there will be build failures. I will comment once all the PRs are ready.

manstis:
+1 to merge.

PR #178, opened 2016-04-14
[GUVNOR-2410] disable UF git/ssh daemons for unit testing

psiroky:
Similar change already done directly in UF.

This will solve all those issues with port binding when running parallel builds.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -384,12 +384,12 @@
               <exclude>**/*IntegrationTest.java</exclude>
             </excludes>
             <argLine>-Xmx1024m -Dfile.encoding=UTF-8</argLine>
-            <systemProperties>
-              <property>
-                <name>apple.awt.UIElement</name>
-                <value>true</value>
-              </property>
-            </systemProperties>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -384,12 +384,12 @@
...
             </excludes>
             <argLine>-Xmx1024m -Dfile.encoding=UTF-8</argLine>
-            <systemProperties>
-              <property>
-                <name>apple.awt.UIElement</name>
-                <value>true</value>
-              </property>
-            </systemProperties>
+            <systemPropertyVariables>
+              <apple.awt.UIElement>true</apple.awt.UIElement>
```
manstis: @psiroky What else are you sneaking in?

psiroky: Nothing really. `<systemProperties>` is deprecated (see https://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#systemProperties) so I just replaced it with `<systemPropertyVariables>`.

manstis:
+1 to merge.

PR #180, opened 2016-04-18
[GUVNOR-2410] disable UF git/ssh daemons for unit testing

psiroky:
Backport of #178. Touches only tests and makes them more reliable in case of parallel build.


manstis:
Gets a +1 from me..

csadilek:
We're already gotten the OK for test-only improvements to go in. So, merging... ;)

PR #181, opened 2016-04-18
upgraded to 6.4.1-SNAPSHOT

kiereleaseuser:


psiroky:
I will start merging these later today/tomorrow as I also need to deploy these new snapshots directly. Otherwise it will take a several hours for jenkins to catch up and we would have lots of false negatives.

ge0ffrey:
+1

PR #182, opened 2016-04-19
upgraded thoughtworks.xstream and springframework

mbiarnes:


kie-platform-bom/pom.xml
```diff
@@ -641,6 +647,13 @@
       </dependency>
 
       <dependency>
+        <groupId>org.springframework</groupId>
+        <artifactId>spring-framework-bom</artifactId>
+        <version>${version.org.springframework}</version>
+        <type>pom</type>
```
psiroky: We also need `<scope>import</scope>`. And I believe that the spring-bom needs to go _before_ the ip-bom, otherwise Maven will use the version from ip-bom (it simply takes the first place the version is, since both versions are provided via imported bom).

mbiarnes:
will close this PR and generate a new one

PR #183, opened 2016-04-21
[DROOLS-1129] use consistent Bundle-SymbolicNames in both binary and …

psiroky:
…sources jars

@mariofusco please take a look. It was in the end much easier than I expected, so I applied the changes right away.


psiroky:
Merging as the other PRs contain +1.

PR #184, opened 2016-04-22
[DROOLS-1137] better granularity for imported BOMs

psiroky:
@ge0ffrey please take a look. This is what we discussed ~2 months back and I finally got some time to implement it.

I removed the `kie-platform-bom` completely. Not sure if it is a good thing or not. It could be useful outside of KIE projects. Will need to think a bit more about that.

The `kie-parent-with-dependencies` should stay just temporarily until I migrate all the other top-level POMs. But I want to do it one at a time, so need to keep the `-with-dependencies` to keep everything building in the mean time.


psiroky:
Please don't merge this yet. I would like to get a review from @ge0ffrey and fix the possible shortcomings.

ge0ffrey:
kie-platform-bom could be used by outsiders, but they shouldn't and I doubt they do (it's not documented AFAIK - you might want to grep the docs for it - or check maven repo stats to see if anything depends on kie-platform-bom). So removing it for 7.0 should be ok.

Not sure if I like the name kie-third-party-bom, but I can't think of anything better.

+1

psiroky:
Yeah, I could not think of better name either. We can always change this in future if we come up with something better.

PR #187, opened 2016-04-27
Use canonical URLs for Maven deploy repositories

psiroky:
This fixes/workarounds the current issues with SNAPSHOT deploys.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -156,6 +156,12 @@
     <latestReleasedVersionFromThisBranch>notYetReleased</latestReleasedVersionFromThisBranch>
     <!-- Version of the KIE Workbench application. Shown for example in About dialog. Will be usually overridden by productisation. -->
     <version.org.kie.workbench.app>${version.org.kie}</version.org.kie.workbench.app>
+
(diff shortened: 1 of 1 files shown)


ge0ffrey:
Is the jbossorg team (David/Vladimir/Vlastimil) ok with this? Do they intend to support that URL forever?
If not, this will break the reproducible of our builds.

psiroky:
They are suggesting it as a workaround. Not sure if they intend to support it forever, will try to find out.

We could always revert back to the old URLs. Currently the only other option I see is to manually fix ~50 Jenkins jobs and adding the property there, which is something I would really like to avoid).

psiroky:
Quoting David Hladký:
`Unless we rename the whole repository.jboss.org we will not rename origin-`

ge0ffrey:
+1 (if the rename the whole, we lose build reproducibility too anyway).

I presume this doesn't affect our local ~/.m2/settings.xml files which might contain a username/password for the deploy servers?

psiroky:
Correct, the server IDs are still the same (jboss-releases-repository and jboss-snapshots-repository), so this change does not affet ~/.m2/settings.xml.

PR #189, opened 2016-04-28
- avoiding JNDI errors when testing with uberfire code

mrietveld:
[This use](https://github.com/uberfire/uberfire/blob/master/uberfire-commons/src/main/java/org/uberfire/commons/async/SimpleAsyncExecutorService.java#L75) of the `InitialContext` is causing errors in my tests, and `simple-jndi` (scope:test) fixes them. :) 

See for more info: http://fandry.blogspot.nl/2011/03/junit-based-integration-testing-with.html

### DIFF
--- a/kie-third-party-bom/pom.xml
+++ b/kie-third-party-bom/pom.xml
@@ -742,6 +742,12 @@
         <version>${version.com.github.detro}</version>
       </dependency>
 
+      <dependency>
+        <groupId>simple-jndi</groupId>
+        <artifactId>simple-jndi</artifactId>
+        <version>${version.simple-jndi}</version>
+        <scope>test</scope>
+      </dependency>
     </dependencies>
 
   </dependencyManagement>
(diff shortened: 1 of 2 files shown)


kie-third-party-bom/pom.xml
```diff
@@ -742,6 +742,18 @@
         <version>${version.com.github.detro}</version>
       </dependency>
 
+      <dependency>
+        <groupId>com.github.detro</groupId>
+        <artifactId>phantomjsdriver</artifactId>
+        <version>${version.com.github.detro}</version>
+      </dependency>
```
psiroky: This seems to be a duplicated dependency? (see the one above)

kie-third-party-bom/pom.xml
```diff
@@ -742,6 +742,18 @@
...
+
+      <dependency>
+        <groupId>simple-jndi</groupId>
+        <artifactId>simple-jndi</artifactId>
+        <version>0.11.4.1</version>
```
psiroky: I guess this got forgotten -- it should use the version property.

kie-third-party-bom/pom.xml
mrietveld: Thanks, will remove these.

mrietveld:
@psiroky Apologies -- the PR has been fixed.

psiroky:
+1

PR #190, opened 2016-05-02
upgrade ironjacamar to align with JBEAP 7.0.0.ER7

mbiarnes:


psiroky:
I believe this is no longer needed. IronJacamar was used by `dashboard-builder` which was recently removed from the build. @mbiarnes could you please check that?

PR #194, opened 2016-05-06
JBPM-5109 - KIE Login Module to capture user credentials upon login t…

mswiderski:
…o use for kie server interaction - added kie security module

this is as part of JBPM-5071 - Clean up and remove the process runtimes from the workbench

to ease collaboration some work that does not affect other components that forked would be good to be placed on master as they are functional as standalone

### DIFF
--- a/kie-bom/pom.xml
+++ b/kie-bom/pom.xml
@@ -156,6 +156,19 @@
         <classifier>sources</classifier>
       </dependency>
 
+      <!-- KIE Security -->
+      <dependency>
+        <groupId>org.kie</groupId>
+        <artifactId>kie-security</artifactId>
+        <version>${version.org.drools.droolsjbpm-integration}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.kie</groupId>
+        <artifactId>kie-security</artifactId>
+        <version>${version.org.drools.droolsjbpm-integration}</version>
+        <classifier>sources</classifier>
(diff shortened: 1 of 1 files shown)


kie-bom/pom.xml
```diff
@@ -157,6 +157,19 @@
         <classifier>sources</classifier>
       </dependency>
 
+      <!-- KIE Security -->
+      <dependency>
+        <groupId>org.drools</groupId>
```
psiroky: Would it be possible to use groupId `org.kie`? Otherwise I guess we should move the module into drools-bom (to follow  the groupId convention).

mswiderski: if that won't be an issues to have it org.kie in droolsjbpm-integration repository....

psiroky: I believe that is fine. There are already other modules with `org.kie` groupId as well...

mswiderski: ok, in that case I'll fix it and use org.kie

mswiderski:
@psiroky change groupId to org.kie

psiroky:
@mswiderski thanks. Should I merge this or wait for the other PR (https://github.com/droolsjbpm/droolsjbpm-integration/pull/465) to get reviewed first?

mswiderski:
@psiroky please leave it for now so the other PR gets reviewed

PR #197, opened 2016-05-10
maven-failsafe-plugin: allow test classes to end also with "IT"

psiroky:
* IntegrationTest is just too long. IT is well established shortcut for
   integration tests so makes sense to support it


mrietveld:
+1

ge0ffrey:
+0

PR #198, opened 2016-05-10
[PLANNER-575] war+ear plugins: upgrade plexus-archiver to 3.0.3 to support JDK9

psiroky:


ge0ffrey:
+1 (We might want to add a comment to indicate why we're adding the workaround, so it's clear when we can delete the workaround once we upgrade to a newer jar/war/ear plugin version.

PR #205, opened 2016-05-19
upgrade jackson-core

mbiarnes:


psiroky:
We should definitely upgrade all jackson deps, otherwise we may encounter incompatibility issues. See the list of the artifacts in ip-bom: https://github.com/jboss-integration/jboss-integration-platform-bom/blob/master/ip-bom/pom.xml#L173

mbiarnes:
@psiroky: OK - I will add the misssing artfacts to kie-third-party pom

mbiarnes:
@psiroky done

psiroky:
Sorry for nit picking, but it would be great to add a comment explaining why we are overriding the version coming from IP BOM. Otherwise few months from now we will be wondering why there is a override and if we could possibly downgrade to the version in IP BOM.

Something like: "Jackson 2.6.2+ required by Dashbuilder's ElasticSearch module. This override can be removed once IP BOM upgrades as well."

romartin:
@psiroky @mbiarnes @dgutierr  Hi guys. This the "why" for this PR: "Current Dashbuilder's ElasticSearch module depends on ELS 1.7.X, which depends on an old version of Lucene that is incompatible with our IP BOM. In order to upgrade the ElasticSearch module to a supported Lucene version, we need to upgrade to ELS 2.X, which requires jackson-core 2.6.2+ at same time. If we don't provide the jackson-core 2.6.2, the ELS .2.x either compiles, as there are new classes in the 2.6.2+ version. So we need the 2.6.2+ in order to use latest ELS 2.X"

romartin:
@psiroky @mbiarnes @dgutierr 

Hey guys, I have completed the upgrade to ELS 2.x and all tests are passing when using my local ELS instance on my laptop. But the default test suite for the ELS connnector tests in dashbuilder is based on running a local ELS instance, firing the test executions and finally stopping the instance. But now I realized that for running the ELS server instance it needs another dependency upgrade :(

-> we are using commons-cli 1.2 and running the ELS server for the tests requires 1.3.1

So another upgrade required here. You think can be done on IP BOM or in our kie-parent? It's important to note that we only need it for test scope.

psiroky:
Merged by https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/commit/6f11bbe635748be33f3c614c61dc97657b213d85

PR #206, opened 2016-05-23
override of commons-clie

mbiarnes:


mbiarnes:
@psiroky: can you remove the commented line:
 ?

psiroky:
Yes, we definitely should. Although, after we merge #207 we won't need that change in depMgmt. It will be enough to just override the property.

psiroky:
@mbiarnes could you please fix this one? It should be enough to just override the version property (and then please verify this via effective-pom).

Not directly related to this PR: just minor thing, but it would be great to use your own github user for these version changing commits (instead of the kiereleaseuser). It is then easier to pin point the exact person who did the change in case of any further questions after it was committed.

psiroky:
Merged by https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/commit/68ab22d071a16efb8afbe4a7f5d2ad61c0f92c75

PR #207, opened 2016-05-23
Inherit directly from IP BOM for easier dependency overrides

psiroky:
* kie-parent now contains all the 3rd party dependency versions.
   It should _not_ import any UI-related BOMs (e.g. errai-bom or
   uf-bom) as non-UI repos don't need them (and they will use
   the kie-parent as well)

 * created kie-user-bom-parent to share the common config for
   user BOMs. This parent inherints directly from ip-parent to
   avoid the dependency imports via kie-parent (user BOMs
   should only declare "our" groupId dependencies, not all the
   3rd party ones. There is kie-platfrom-bom for that).

 * re-introduced kie-platfrom-bom as it comes handy for
   some downstream projects (e.g. tests)

 * fixed several POM metadata tags like SCM, URL, mailing lists

 * using the property ${version.org.kie} consistently since to make
   it more obvious that all KIE projects have the same
   release cycle (the other version properties like version.org.drools
   or version.org.jbpm should be slowly retired, together with fixing
   all the possible downstream usages)

@mbiarnes please take a look. These are the same changes as were recently done for both UberFire and Dashbuilder repos.


mbiarnes:
+1 - great change!

PR #208, opened 2016-05-24
Fix update-versions-all after recent kie-parent changes

psiroky:
* we need to separately change the versions for the user BOMs since they no longer inherit from the top-level kie-parent

@mbiarnes please take a look


mbiarnes:
+1 checked locally and it works perfect

PR #210, opened 2016-05-24
Remove dashboard-builder and jbpm-dashbuilder from the repository-list.txt

psiroky:
* these repos will never be released as part of 7.x series,
   so there is no need to keep building them


psiroky:
Replaced by https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/213 (since I can't easily update this branch).

PR #211, opened 2016-05-25
Remove no longer needed parts of depMgmt

psiroky:
* the override is either no longer necessary,
   or it can be done only via version property

@mbiarnes please take a look


psiroky:
Jenkins, execute full downstream build

mbiarnes:
+1 - all this removed dependencies are in jboss-ip-bom

PR #212, opened 2016-05-25
GUVNOR-2181: Guided Decision Table Editor: Add support for linked tables

manstis:
See https://issues.jboss.org/browse/GUVNOR-2181 and referenced BPSMPL.

@psiroky @romartin Is working on a "real" Lienzo Mockito release and GAV.... this will be updated.

This is Phase 1, replacing the existing Grid Widget.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -71,7 +71,8 @@
     <!-- ################################################################################ -->
     <version.avalon-framework>4.1.4</version.avalon-framework>
     <version.batik>1.6.1-1</version.batik>
-    <version.com.ahome-it.lienzo-core>2.0.192-RC1</version.com.ahome-it.lienzo-core>
+    <version.com.ahome-it.lienzo-core>2.0.270-RELEASE</version.com.ahome-it.lienzo-core>
+    <version.com.ahome-it.lienzo-tests>1.0.0-RC1</version.com.ahome-it.lienzo-tests>
     <version.com.android-dx>1.10</version.com.android-dx>
     <version.com.github.detro>1.2.0</version.com.github.detro>
     <version.com.github.tomakehurst.wiremock>1.53</version.com.github.tomakehurst.wiremock>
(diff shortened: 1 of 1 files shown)


manstis:
Part of an ensemble:

https://github.com/uberfire/uberfire/pull/379
https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/212
https://github.com/droolsjbpm/kie-wb-common/pull/368
https://github.com/droolsjbpm/drools-wb/pull/185
https://github.com/droolsjbpm/kie-wb-distributions/pull/293

manstis:
@mariofusco @psiroky This is good to go now.

pom.xml
```diff
@@ -71,11 +71,14 @@
...
     <version.batik>1.6.1-1</version.batik>
-    <version.com.ahome-it.lienzo-core>2.0.192-RC1</version.com.ahome-it.lienzo-core>
+    <version.com.ahome-it.lienzo-core>2.0.270-RELEASE</version.com.ahome-it.lienzo-core>
+    <version.com.ahome-it.lienzo-tests>1.0.0-RC1</version.com.ahome-it.lienzo-tests>
     <version.com.android-dx>1.10</version.com.android-dx>
     <version.com.github.detro>1.2.0</version.com.github.detro>
     <version.com.github.tomakehurst.wiremock>1.53</version.com.github.tomakehurst.wiremock>
     <version.com.google.android>4.1.1.4</version.com.google.android>
+    <!-- Override the version in ip-bom until https://github.com/jboss-integration/jboss-integration-platform-bom/pull/310 is merged -->
+    <version.com.google.gwt.gwtmockito>1.1.6</version.com.google.gwt.gwtmockito>
```
psiroky: This override should not be necessary anymore as we are now using ip-bom 7.0.0CR3 which defines gwtmockito in version 1.1.6.

pom.xml
```diff
@@ -1723,6 +1726,13 @@
       </dependency>
 
       <dependency>
+        <groupId>com.ahome-it</groupId>
+        <artifactId>lienzo-tests</artifactId>
+        <version>${version.com.ahome-it.lienzo-tests}</version>
+        <scope>test</scope>
```
psiroky: Nit picking, but we should not define scope in dependencyManagement. Modules depending on this one are in charge of deciding which scope they need. This would work for most of the case (or maybe all of them), but makes it slightly harder to understand since it changes the default scope from `compile` to `test` (e.g. if the module depending on this artifact does not specify scope, it would be `test` instead of the otherwise defaulted `compile`)

manstis:
@psiroky Thanks for your comments. 

I've modified accordingly (and similarly for https://github.com/uberfire/uberfire/issues/379)

psiroky:
Thanks @manstis! +1 to merge.

PR #213, opened 2016-05-27
Remove dashboard-builder and jbpm-dashbuilder from the repository-lis…

psiroky:
…t.txt

 * these repos will never be released as part of 7.x series,
   so there is no need to keep building them

@csadilek, @dgutierr thoughts?


csadilek:
Fine with me. We don't want to release these projects as part of 7 as that would mean keeping them for many more years. So, I am OK with removing them now. @dgutierr what do you think?

jhrcek:
+1 If this PR were merged into 7.0 branch it would mean that we could delete our large dashbuilder selenium test suite from our 7.0 branch. I'm watching this PR with great interest :-)

jhrcek:
@dgutierr could you comment on this, please?

dgutierr:
+1

psiroky:
Please don't merge yet, I need to double check what fixes need to be done regarding our automation. I will merge the PR as soon as I am confident it won't break anything (like PR builds).

PR #222, opened 2016-06-18
Fixes after IP BOM 7.0.0.CR4 upgrade

psiroky:
* downgrade surefire&failsafe from 2.19.1 to 2.18.1
   as 2.19.1 is buggy

 * remove some config which is already brought in by
   jboss-parent


psiroky:
All PRs mentioned in https://github.com/errai/errai/pull/180 need to be merged together, at the same time.

csadilek:
+1

PR #228, opened 2016-06-28
Add resteasy-client dependency

livthomas:
I have added resteasy-client dependency which is needed for the REST client in kie-wb-distributions/kie-wb-tests/kie-wb-tests-rest. I will send a pull request to jboss-integration-platform-bom but in the meantime we need to have it at least here.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1649,6 +1649,12 @@
         <version>${version.simple-jndi}</version>
         <scope>test</scope>
       </dependency>
+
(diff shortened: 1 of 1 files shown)


psiroky:
+1, merged as https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/commit/43dfd9f4e15195cc64005170ecb3ca562fa8c436

PR #234, opened 2016-07-06
GUVNOR-2600: Expose Dynamic Bean Validation through DataModelOracle

manstis:
See https://issues.jboss.org/browse/GUVNOR-2600

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -126,6 +126,10 @@
     <version.simple-jndi>0.11.4.1</version.simple-jndi>
     <version.org.asciidoctor.pdf>1.5.0-alpha.11</version.org.asciidoctor.pdf>
 
+    <!-- Required for Errai Dynamic Bean validation  -->
+    <version.org.hibernate.hibernate-validator>4.1.0.Final</version.org.hibernate.hibernate-validator>
+    <version.javax.validation.validation-api>1.0.0.GA</version.javax.validation.validation-api>
+
     <!-- In community builds productized is false, in product builds it's true to enable branding changes -->
     <org.kie.productized>false</org.kie.productized>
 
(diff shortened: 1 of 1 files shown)


manstis:
Part of an ensemble:

https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/234
https://github.com/droolsjbpm/drools/pull/828
https://github.com/droolsjbpm/kie-wb-common/pull/402
https://github.com/droolsjbpm/drools-wb/pull/236

csadilek:
+1

psiroky:
I am not sure how will this play out when we upgrade to Hibernate 5.x.

@manstis, @csadilek: do I understand correctly that we only need to override the hibernate-validator version for GWT compilation? Backend (e.r.g WEB-INF/lib) should use whatever version we are aligned to (soon to be 5.x)?

manstis:
@psiroky @csadilek That's a great question.. and I hope a simple answer! I believe you are correct. Where we actually use the hibernate 4.1.0.Final dependency for GWT it is "provided" scope and used for the different validators source code to compile into JavaScript... so after the webapp (GWT) compilation the dependency is no longer needed. IDK if GWT2.8+ Bean Validation will work with hibernate 5.x..  if it does not we may have an issue compiling kie-(drools)-wb as GWT will need 4.1.0.Final on the classpath at compile time and (presumably) drools/jbpm persistence 5.x.

psiroky:
@manstis yeah, that's exactly why I was looking into this -- the possible classpath clash for -webapps. 

@csadilek mentioned that GWT 2.8+ will _not_ work with Hibernate 5.x, in fact the whole bean validation thing is being deprecated in GWT (and I guess will eventually move to Errai, or some other third party library).

csadilek:
@manstis @psiroky yes, exactly. If we can't use separate client and server-side dependencies there (I believe that should be possible), we will have to upgrade the validation GWT module and move it to Errai.

PR #240, opened 2016-07-14
UF-305 Add version management for new theme modules

cristianonicolai:


manstis:
+1 to merge (on face value).

@cristianonicolai Is this PR to remain unaltered? There was some talk about the number of Patternfly related GWT modules being inherited and I'm unclear as to whether anything will be changed? Perhaps ping BSIG or somewhere when the PR review has stabilised?

PR #243, opened 2016-07-20
added scripts for pushing release branches

mbiarnes:


mbiarnes:
@psiroky 
can you please check this and merge it, so I can do tests on the "real" environment and see if everything works.

script/release/01.KIE-releases_push-release-branches.sh
```diff
@@ -0,0 +1,84 @@
+BASE_BRANCH=6.4.x
+
+# clone the build-bootstrap that contains the other build scripts
+if [ "$SOURCE" == "community-branch" ]; then
+     
+   # clone droolsjbm-build-bootstrap branch from droolsjbpm
+   git clone git@github.com:droolsjbpm/droolsjbpm-build-bootstrap.git --branch $BASE_BRANCH
```
psiroky: Since this script is already in the droolsjbpm-build-bootstrap git repo, this clone seem unnecessary as the repo already need to exist in order to have access to this script. Or am I missing something?

script/release/01.push-release-branches-DSL.txt
```diff
@@ -0,0 +1,66 @@
+def cloneScript_01 =
```
psiroky: @mbiarnes did you intend to push this DSL file? It uses hard-coded paths to workspaces, etc.

mbiarnes: @psiroky 
I think it could be a benefit if we have the DSL script stored somewhere in order not having to redefine it when this Job get lost.
I think it is a good idea to change all hard-coded path to a placeholder,

mrietveld: Another possibility is to do the following: 

- add a properties file in this directory (on your laptop) that contains these values
- add a .gitignore file that ignores the properties file
- add code to the script that reads in the properties file and assigns/uses the values..

psiroky: @mbiarnes I completely agree we need to store the DSL files in git repo. We just need to avoid hard-coding the paths (to make it as much resilient and stable as possible). I'll take a closer look at this today or tomorrow to see how we can easily do that.

mbiarnes:
I will close this PR since there will be new refactored scripts and a new PR will be raised

PR #245, opened 2016-07-28
Upgrade Cargo from 1.4.19 to 1.5.0

psiroky:
* 1.5.0 brings support for EAP 6.4.7+

cc @mswiderski, @sutaakar


mswiderski:
as long as it does work as expected +1, though I let @sutaakar to have final word as he's cargo expert

sutaakar:
+1

PR #247, opened 2016-08-01
moved 6.5.x to jdk version 1.8

mbiarnes:


psiroky:
@mbiarnes sorry to mention this only now, but I think it would be good to include some more info. We should be clear about the support of JDK 6. It is too easy to think that we support only JDK8 based on this property, which could create certain confusion.

Something like "Important: JDK8 is needed to build the projects (to support also WildFly 10), but we still _do_ support JDK6/JDK7 for all backend modules. "

psiroky:
@ge0ffrey let me try to explain this little bit more. We will use JDK8 to build the projects, but we will still keep the JDK6 compatibility for all backend stuff. We did something very similar with 6.4.x as well. Because of some GWT lib, we had to build with JDK7, but we kept the JDK6 compatibility.

IP BOM defines source/target versions for maven-compiler-plugin, see https://github.com/jboss-integration/jboss-integration-platform-bom/blob/6.0.x/pom.xml#L95. They are still "1.6" and we are not overwriting those in our projects.

The `jdk.min.version` property is only used by enforcer to enforcer the JDK version during build.

ge0ffrey:
+0 in that case (I retract my veto). Thanks for the enlightenment @psiroky (didn't notice the diff between jdk.min.version and maven.compiler.target).

ge0ffrey:
By enforcing JDK 8 during the build, we can't validate that no since 1.7 or since 1.8 code is used.
In JDK 9, we'll be able to use -release 9 (which replaces -source 9 and -target 9 too) to verify that without having to use an earlier JDK. IIRC, you do have some sort of "animal" plugin to verify that - is that one water proof?

psiroky:
Yes, we do use animal-sniffer-plugin to detect newer Java API usage. https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/blob/6.5.x/pom.xml#L316

psiroky:
And yes, this is definitely not straightforward. That's why I asked @mbiarnes to put a clear comment into the POM to avoid any further confusion.

mbiarnes:
@psiroky DONE

psiroky:
@mbiarnes thanks, although I was thinking about adding this directly into the pom.xml where it would be more visible comparing to the commit message.

Anyway, I believe we should merge this together with other repos (uf, dashbuilder). @mbiarnes could you please create PRs also for those?

mbiarnes:
@psiroky Added comment to pom. 
Do you mean with "merge this together with other repos" to edit jdk.min.version in the poms of uberfire/uberfire-extensions and dashbuilder?

psiroky:
@mbiarnes yes, if we decide to use this (like we probably should), we should be consistent across all the repos.

psiroky:
+1 to merge

PR #250, opened 2016-08-05
Upgrade to GWT 2.8.0-rc1

csadilek:
Most external dependencies that used to be inside the gwt-dev.jar have now been unbundled. So we can get rid of the enforcer exception and can now manage exclusions using the standard maven mechanism.


manstis:
+1 to merge.

PR #251, opened 2016-08-12
Use property to easily enable/disable checkstyle

psiroky:
* the failOnValidation is false by default since
   most of the repos would fail on that. One repo
   at the time, we need to fix them, override
   the property there (temporarily) and once all
   repos comply, we can also change the property
   value here

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -156,6 +156,8 @@
     <!-- Surefire version 2.19.1 coming from parent is buggy and drools-osgi tests in droolsjbpm-integration are failing because of that. -->
     <version.surefire.plugin>2.18.1</version.surefire.plugin>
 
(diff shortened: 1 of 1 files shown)


csadilek:
+1

ge0ffrey:
To get them to comply, it can be done quickly with regex in IntelliJ:
Replace "\t" with "    " in all "*.java" files.
I actually did that for all repo's in 2010, but as you can imagine enforcing is the only way to make it so.

ge0ffrey:
Btw, are we using .gitattributes on all repo's already to enforce line endings without any pain whatsoever (even for contributors sending in a PR from windows)? I use it for optaplanner and it's a dream:
  https://github.com/droolsjbpm/optaplanner/blob/master/.gitattributes
I created a jira for that a long time ago.

PR #254, opened 2016-08-18
Upgrade Cargo from 1.4.19 to 1.5.0 to support latest EAP 6.4.x

psiroky:


psiroky:
@mswiderski FYI. This should be safe upgrade I believe. @sutaakar WDYT?

mswiderski:
@psiroky could we name the branch same as https://github.com/droolsjbpm/droolsjbpm-integration/pull/580 so we get direct verification when retriggering the droolsjbpm-integration PR?

psiroky:
@mswiderski yes I can do that, but it needs new PR. Will create one in a sec (I did try locally, so should be fine. But I agree its better run Jenkins build as well).

psiroky:
Replaced by https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/255

PR #255, opened 2016-08-18
Upgrade Cargo from 1.4.19 to 1.5.0 to support latest EAP 6.4.x

psiroky:


sutaakar:
+1, used just for tests, shouldn't make any problem

PR #256, opened 2016-08-18
gwt-maven-plugin: don't print entire java command on error

psiroky:
* the java command also contains the whole classpath, which
   in our case means literally thousands of characters. Leaving
   out the command makes the Maven build log much cleaner. It is
   now also much easier to find the actual root cause of the
   failed GWT compilation

The quest for clean GWT build logs in almost complete! Only one PR away :)

CC @csadilek


csadilek:
+1

PR #258, opened 2016-08-19
added scripts for KIE releases

mbiarnes:
scripts for community releases or prod tags/releases of 6.5.x with DSL plugin

### DIFF
--- a/script/dsl-scripts/KIE_Release.sh
+++ b/script/dsl-scripts/KIE_Release.sh
@@ -1,0 +1,628 @@
+def releasePrepare =
+"""
+sh /home/jenkins/workspace/DSL-KIE-Release-6.5.x/script/release/01.KIE_releasePrepare.sh
+"""
+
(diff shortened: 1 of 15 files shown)


psiroky:
I'll look at this as soon as I can (likely on Monday).

script/dsl-scripts/KIE_Release.sh
```diff
@@ -0,0 +1,628 @@
+def releasePrepare =
+"""
+sh /home/jenkins/workspace/DSL-KIE-Release-6.5.x/script/release/01.KIE_releasePrepare.sh
```
psiroky: We really should not use hardcoded paths here. It makes the scripts hard to use on different environments (e.g. locally when testing the changes). Could we just use the relative path `script/release/<name>`?

mbiarnes:
I will close this PR since there will be new re-factored scripts and a new PR will be raised

PR #261, opened 2016-08-23
[6.5.x] Use property to easily enable/disable checkstyle

psiroky:
* the failOnValidation is false by default since
   most of the repos would fail on that. One repo
   at the time, we need to fix them, override
   the property there (temporarily) and once all
   repos comply, we can also change the property
   value here

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -132,7 +132,8 @@
     <!-- Override6.1.1 is the last version supporting Java 6. We need to be able to run certain modules
          (e.g. jbpm-test-coverage) with Java 6. -->
     <version.checkstyle>6.1.1</version.checkstyle>
-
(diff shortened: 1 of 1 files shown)


ge0ffrey:
+1 It's a pity that the checkstyle plugin doesn't seem to expose a command line property to overwrite it from the command line out of the box, like it does with other properties, for example -Dcheckstyle.enable.severity.summary=true
https://maven.apache.org/plugins/maven-checkstyle-plugin/checkstyle-mojo.html

psiroky:
There is a property `checkstyle.failOnValidation` which you can use from command line. However, in case one directly uses in the plugin config `<failOnViolation>false</failOnViolation>`, the property is then ignored. I have seen this with other plugins as well, but not sure if this is actually the expected behavior.

PR #267, opened 2016-08-30
RHBRMS-2525: Business-central startup error: jGIT: Cannot run program 'bash'

manstis:
See https://issues.jboss.org/browse/RHBRMS-2525

This overrides the version in ip-bom until we move to a version containing https://github.com/jboss-integration/jboss-integration-platform-bom/pull/327


manstis:
@psiroky Hi.. I think this closes the loose ends (until we can use a newer ip-bom). WDYT?

psiroky:
I believe we should override the version also in uberfire to make sure the jgit integration still works with the latest version.

manstis:
@psiroky Good catch!

See https://github.com/uberfire/uberfire/pull/477 too.

psiroky:
+1 to merge (together with the UF one)

PR #269, opened 2016-08-31
Allow custom repo lists for git-all and mvn-all scripts

psiroky:
@mbiarnes could you please review this?

### DIFF
--- a/script/git-all.sh
+++ b/script/git-all.sh
@@ -22,14 +22,34 @@ initializeWorkingDirAndScriptDir() {
 initializeWorkingDirAndScriptDir
 droolsjbpmOrganizationDir="$scriptDir/../.."
 
-if [ $# = 0 ] ; then
+# default repository list is stored in the repository-list.txt file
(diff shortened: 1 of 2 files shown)


ge0ffrey:
this is useful indeed. Now I have separate checkouts for "all repos", "just drools" (includes knowledge, docs), "just-wb" (all wb repo's), etc

psiroky:
In future I would also like to add an option to specify own repository list as a file (e.g. --repository-list-file=)

Also, one other thing that looks useful it something like Maven's `-rf` (run from), so that if the build fails in the middle you can easily just say to start building from the repository that failed (instead of having to rebuild all the previous ones as well).

etirelli:
Quite helpful! Thank you Petr!

ge0ffrey:
or -rt (--run-to). When working on the engines, it's handy to say "-rt optaplanner" for me or "-rt drools" for @mariofusco

kiereleaseuser:
+1

PR #270, opened 2016-08-31
RHBRMS-2552 Define a property for the latest Kie release.

AntStephenson:
Define a property for the latest Kie release, to be used in
droolsjbpm-knowledge/kie-api when checking API compatibility,
this allows productization to override this.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -184,6 +184,9 @@
          directly should help to avoid some of the issues with deploying artifacts. -->
     <jboss.releases.repo.url>https://origin-repository.jboss.org/nexus/service/local/staging/deploy/maven2/</jboss.releases.repo.url>
     <jboss.snapshots.repo.url>https://origin-repository.jboss.org/nexus/content/repositories/snapshots/</jboss.snapshots.repo.url>
+
+    <!-- Latest release to be used by api-compatibility-check to check backwards compat of the Kie API. -->
+    <version.org.kie.latestFinal.release>6.4.0.Final</version.org.kie.latestFinal.release>
   </properties>
 
   <repositories>


ge0ffrey:
This impacts the release procedure I believe, so it should probably include changes on this document:
  https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/blob/master/RELEASE-README.md

ge0ffrey:
Has there been any research to see if the previous final release can be deducted automatically? Maybe with maven-versions-plugin etc? Anything that lengthens the release procedure brings us further from the "push button to release" dream.

psiroky:
I am not sure if `version.org.kie.latest.release` is the best name, since it may not be latest release -- e.g. for current master the latest release is 7.0.0.Beta1.

Maybe we could call it `version.org.kie.latest.final.release`? (assuming that is what we want)

psiroky:
Deducing it automatically would be great, but I am currently not aware of easy way to do that.

AntStephenson:
Hi guys,

Thanks for the comments, @ge0ffrey I'll have to investigate this tomorrow and post my findings here. @psiroky this makes sense to me, thanks.

ryanzhang:
+1 on version.org.kie.latest.final.release

This is for replacing the hardcode version defined in droolsjbpm-knowledge (the hardcode version is defined inside revapi maven plugin)
It work with  https://github.com/droolsjbpm/droolsjbpm-knowledge/pull/165

@ge0ffrey Can you please specify what change are expected in the release docs in upstream?

ge0ffrey:
@ryanzhang I would guess that the value of this property needs to change on every Beta1 release. Otherwise it will just go stale in community.

Anything that needs to happen during a release - and isn't done automatically - should be clearly mentioned in the release procedure - so when MichaelB or anyone else does a community release, it is taken care of. The vice versa is also true: if something is not mentioned in the release procedure, it must not be done during a release, because we believe in release consistency.

ge0ffrey:
About the naming: let's keep it consistent with drools-website, optaplanner-website and jbpm-website. Those already use the following names:
- latest: the latest release, for example 7.0.0.Beta1
- latestFinal: the latest Final release, for example 6.4.0.Final

See for example https://github.com/droolsjbpm/drools-website/blob/master/_config/pom.yml

So I guess `version.org.kie.latest.final.release` is fine (same wording), but `version.org.kie.latestFinal.release` might be a bit more consistent (same wording and same camelCase). It's bikeshedding, I guess - you guys decide :)

ryanzhang:
OK, agree with version.org.kie.latestFinal.release since it's more consistent to the existing practise.

Regarding to properties changes, I think it won't change and won't go stale for kie 6.5.x Beta release. 
It means the latest Final release. ie 6.4.0.Final. 
It is used to replace the hardcode version in https://github.com/droolsjbpm/droolsjbpm-knowledge/pull/165/files

IIUC, it won't change for every milestone release. It's used to check the api compatiblity between major release.

ryanzhang:
@ge0ffrey @psiroky Could you please merge this PR? or is there  anything else need to modify?
We target to fix this for ER2.
Also https://github.com/droolsjbpm/droolsjbpm-knowledge/pull/165 is related to the same issue. 
Those 2 PR worked together but they are in two different repo.

AntStephenson:
Need to update the property to version.org.kie.latestFinal.release, before merging. I'll do that soon.

ryanzhang:
@AntStephenson I see, Thanks!

ge0ffrey:
+1 to merge after Ant's upcoming change. I presume Petr will do the actual merge as he gatekeeps this repo.

AntStephenson:
I've updated both PRs to reflect the discussion here. Let me know if there's anything else you need me to do!

psiroky:
Thanks @AntStephenson. We will also need to update the release procedure (I can do that). I believe the version needs to change when we branch (and not when we release next Beta1). We need the version updated as soon as Final is released.

So for example when we release 6.5.0.Final this needs to be right away updated on master (where we have 6.4.0.Final right now).

PR #273, opened 2016-09-05
Fix build on Java 8: disable javadoc doclint

psiroky:
* Java 8 javadoc tool is more strict and fails the build.
   This has been already fixed on newer branches

@manstis this should fix the PR builds on 6.3.x branch


manstis:
+1 to merge. Thanks @psiroky

PR #279, opened 2016-09-13
added springframework temporary that overwrites version in jboss-ip-b…

mbiarnes:
…om 6.0.8.Final


mbiarnes:
+1

psiroky:
I am afraid that just overriding the property is not enough for 6.5.x. We also need to copy&paste the respective part of `<dependencyManagement>`. Or upgrade to newer IP BOM which already contains that spring upgrade.

PR #287, opened 2016-10-04
GUVNOR-2709 : Move client side Verification codes to a separate module

Rikkola:
https://issues.jboss.org/browse/GUVNOR-2709
There are 3 PRs in the change bundle:
https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/287 
https://github.com/droolsjbpm/drools-wb/pull/309
https://github.com/droolsjbpm/kie-wb-distributions/pull/369

### DIFF
--- a/drools-bom/pom.xml
+++ b/drools-bom/pom.xml
@@ -473,6 +473,20 @@
         <artifactId>drools-rhq-plugin</artifactId>
         <version>${version.org.kie}</version>
       </dependency>
+
+      <!--Verification-->
+      <dependency>
+        <groupId>org.drools</groupId>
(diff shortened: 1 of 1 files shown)


drools-bom/pom.xml
```diff
@@ -473,6 +473,20 @@
         <artifactId>drools-rhq-plugin</artifactId>
         <version>${version.org.kie}</version>
       </dependency>
+
+      <!--Categories Editor-->
```
manstis: Yeah.. it's probably not the "Categories Editor" now....

Rikkola: Fixed.

PR #288, opened 2016-10-04
git-all,mvn-all: allow spaces as part of arg. lines

psiroky:
@mbiarnes please take a look. This should fix the issue with spaces as part of commit msg (the issue is not related to the Jenkins upgrade).


mbiarnes:
+1 - tested on Jenkins CI

PR #291, opened 2016-10-07
Fix IntelliJ config + change method parameters to align when multiline

ge0ffrey:
The intelliJ config was broken, it failed to import in recent intellij versions.
This fixes that and it's pretty much the same style as it was back then.

It does have one noticeable change: align method parameters when multiline. This is against the original Java coding conventions:
  http://www.oracle.com/technetwork/java/javase/documentation/codeconventions-136091.html#248
but it seems the majority prefers it this way (I 'd prefer to stick to the standard though :)).

### DIFF
--- a/ide-configuration/intellij-configuration/code-style/intellij-code-style_droolsjbpm-java-conventions.xml
+++ b/ide-configuration/intellij-configuration/code-style/intellij-code-style_droolsjbpm-java-conventions.xml
@@ -69,7 +69,6 @@
     <option name="KEEP_BLANK_LINES_IN_CODE" value="1" />
     <option name="KEEP_BLANK_LINES_BEFORE_RBRACE" value="1" />
     <option name="BLANK_LINES_AFTER_CLASS_HEADER" value="1" />
-    <option name="ALIGN_MULTILINE_PARAMETERS" value="false" />
     <option name="ALIGN_MULTILINE_FOR" value="false" />
     <option name="IF_BRACE_FORCE" value="3" />
     <option name="DOWHILE_BRACE_FORCE" value="3" />
--- a/ide-configuration/intellij-configuration/code-style/intellij-code-style_droolsjbpm-java-conventions.xml
+++ b/ide-configuration/intellij-configuration/code-style/intellij-code-style_droolsjbpm-java-conventions.xml
@@ -1,12 +1,11 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<code_scheme name="Drools and jBPM: Java Conventions">
-  <option name="LINE_SEPARATOR" value="&#10;" />
-  <option name="USE_FQ_CLASS_NAMES_IN_JAVADOC" value="false" />
+<code_scheme name="KIE Java Conventions">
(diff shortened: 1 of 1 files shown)


manstis:
Thanks @ge0ffrey 

Personally I also prefer the following to "Wrap always" and "Align when multiline":-

1) "Extends/implements list"
2) "Method declaration parameters"
3) "Method call arguments"

(1) and (2) are purely cosmetic but IMO make it easier to identify super-class/interfaces and overloaded methods when scanning code quickly, (3) IMO makes debugging easier when you're stepping into the invoked method and the passed parameters are derived from other method calls e.g. 
```
m.showMeTheMoney( getAgent(),
                  getPlayer() );
```
There's a few other tweaks I seem to have acquired (compared to this PR).. spaces in method calls and some others, e.g. ```w.smurftastic("pupa");``` vs ```w.smurftastic( "pupa" );```.

I fear nobody else is going to comment on this PR..

psiroky:
I fully agree with @ge0ffrey. I would prefer to stay with the standard when it comes to the align method parameters when multiline

ide-configuration/intellij-configuration/code-style/intellij-code-style_droolsjbpm-java-conventions.xml
```diff
@@ -1,12 +1,11 @@
...
-  <option name="LINE_SEPARATOR" value="&#10;" />
-  <option name="USE_FQ_CLASS_NAMES_IN_JAVADOC" value="false" />
+<code_scheme name="KIE Java Conventions">
+  <option name="LINE_SEPARATOR" value="&#xA;" />
   <option name="CLASS_COUNT_TO_USE_IMPORT_ON_DEMAND" value="99999" />
   <option name="NAMES_COUNT_TO_USE_IMPORT_ON_DEMAND" value="1" />
   <option name="PACKAGES_TO_USE_IMPORT_ON_DEMAND">
     <value>
       <package name="org.junit.Assert" withSubpackages="true" static="true" />
+      <package name="org.mockito.Mockito" withSubpackages="true" static="true" />
```
triceo: Can I ask that AssertJ be added, too? The same as the above, so should share the behavior.

ge0ffrey:
@manstis I fear everybody's going to comment on this PR :) It's easy to have a strong opinion on code style ( https://www.youtube.com/watch?v=SsoOG6ZeyUI ).
When that happens, we 're going to try to design a code style by committee and this PR will never get merged and we'll be stuck with the original broken config file.

psiroky:
I'll merge this so we can at least import the style file. We can (and will) always argue about the nuances later :)

ge0ffrey:
+1 (and it's 100% ok to leave out my second commit there ;)

psiroky:
Merged as https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/commit/dc77c93f9d60507382831612f6ddf0024abba02a and https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/commit/1d35e1062cc735557a2be7177361e988ef728218

PR #296, opened 2016-10-13
[JBPM-5366] remove most of kie-remote deps

psiroky:
* kie-remote-common is still being used by kie-server,
   so can not be removed at this point

Depends on https://github.com/droolsjbpm/droolsjbpm-integration/pull/685


psiroky:
@mswiderski we need to merge this one as well to fix the build

PR #299, opened 2016-10-19
GUVNOR-2722: i18n: Host JSP's do not use correct Locale when the User has configured multiple. 

manstis:
Correct <includes> element.

@mbiarnes Hi, could you please approve?

I've tested with ```mvn zanata:push-module -B``` on ```kie-wb-distributions``` and it was fine.


csadilek:
+1

PR #302, opened 2016-10-24
JBPM-5381: Added Stunner artifacts into KIE BOM that provide the workbench & guvnor integration.

romartin:
See https://issues.jboss.org/browse/JBPM-5381


csadilek:
+1

PR #307, opened 2016-10-31
RHBPMS-4360: upgrade jgit version in order to avoid windows issues on high concurrency situations.

porcelli:
Important note: the jgit upgrade requires Java7+, so this is also removing 1.6 support from the build.


csadilek:
+1

pom.xml
```diff
@@ -416,6 +418,8 @@
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-compiler-plugin</artifactId>
           <configuration>
+            <source>1.7</source>
+            <target>1.7</target>
```
psiroky: I believe it would be better to override the source/target properties coming from parent POM(s) (those are then used by the compiler plugin). Specifically `<maven.compiler.target>1.7</maven.compiler.target>` and `<maven.compiler.source>1.7</maven.compiler.source>`. See https://github.com/jboss-integration/jboss-integration-platform-bom/blob/6.0.x/pom.xml#L94 for more info.

PR #308, opened 2016-11-02
Upgrade to Guava 20.0

csadilek:
@psiroky please ;).

IMO, we can delete our guava fork repo now.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -123,7 +123,7 @@
     <version.com.google.gwt>2.8.0</version.com.google.gwt>
     <!-- TODO Remove this when gson version is updated in ip-bom -->
     <version.com.google.code.gson>2.6.2</version.com.google.code.gson>    
-    <version.com.google.guava>20.0-rc1</version.com.google.guava>
+    <version.com.google.guava>20.0</version.com.google.guava>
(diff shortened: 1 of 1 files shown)


ge0ffrey:
+1 to upgrade, that being said:

Guava has a very aggressive 18 month deprecation-then-remove policy. So I think we need **an inventory of what deprecated guava methods/classes we're now still using in our compilation build**. We can then split up that inventory and send it to the leads. Otherwise we're going to get into trouble when upgrading to version 21 or 22. Who would be best positioned to make that inventory?

psiroky:
@ge0ffrey good point. Not sure who should be responsible for that though...

In ideal case we could look at compiler warnings, but since there are hundreds of warnings already, its gonna be harder to spot the Guava ones.

ge0ffrey:
grep :)

I think we need to look at all deprecated compiler warnings anyway, once a year. I think we need a formal "inventory" event with a number of things we need to do, once per year, with inventory lists posted to bsig with things such as use of deprecated methods, findbugs, dependency tree listings, ... split up on drools/optaplanner/jbpm/dashbuilder/errai/uberfire so the project leads can split them up further to delegate. I don't think they need to be resolved 100% (exceptions define the rule), but reducing them by 80% should be feasible.

PR #312, opened 2016-11-10
Add version for arquillian wildfly container

cristianonicolai:


pom.xml
```diff
@@ -1817,6 +1818,12 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+          <groupId>org.wildfly.arquillian</groupId>
+          <artifactId>wildfly-arquillian-container-managed</artifactId>
+          <version>${version.org.jboss.arquillian.container.wildfly}</version>
```
ge0ffrey: The version name is nonstandard, it should start with org.wildfly.arquillian instead of org.jboss.arquillian, so it is predictable when looking for in the list (which is alphabetically so when looking for org.wildfly we won't find org.jboss). 

I suggest we name it version.org.wildfly.arquillian.container, I named it like that already in optaplanner-webexamples (which will need to discard it's version overwrites)
https://github.com/droolsjbpm/optaplanner/blob/master/optaplanner-webexamples/pom.xml#L25

cristianonicolai:
@ge0ffrey thanks for checking that, changes applied.

ge0ffrey:
Thanks Cristiano. +1 to merge :)

ge0ffrey:
Also cleaned up optaplanner-webexamples accordingly

PR #314, opened 2016-11-13
New Library Perspective (new authoring)

ederign:


manstis:
I suggest merge when the other PRs are ready for merge too.

porcelli:
+1 with same caution that @manstis mentioned

PR #319, opened 2016-11-23
maven-compiler-plugin: disable forking + remove duplicated config

psiroky:
* jboss-parent already defines source, target, warnings and deprecation
 * fork config removed == forking disabled (default is false)
 * meminitial and memmax are only needed when forking
 * compile-only builds are 60 % faster with forking disabled

@mbiarnes, @ge0ffrey I don't know if there was a specific reason to always fork the compiler, do you remember?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -438,20 +438,6 @@
         </plugin>
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
-          <artifactId>maven-compiler-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


psiroky:
Jenkins, execute full downstream build

psiroky:
@mbiarnes, @ge0ffrey could you please take a look at this?

mbiarnes:
+1

PR #323, opened 2016-11-30
maven-compiler-plugin: temporarily enable forking

psiroky:


pom.xml
```diff
@@ -381,6 +381,15 @@
           </executions>
         </plugin>
         <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-compiler-plugin</artifactId>
+          <configuration>
+            <!-- Forking temporarily enabled as it seems to workaround the wrong behavior of UF annotation processor, resulting
+                 in corrupted perspective definitions, etc. This needs to fixed directly in UF and the forking disabled again. -->
+            <fork>true</fork>
```
psiroky: The only way I see is to configure all the workbench modules to override the config, which is quite painful.

I am now looking into other alternatives (e.g. fixing the UF stuff or some advanced plugin config)

psiroky:
Merging this for now to make the builds work. Once I find a better way to fix the issue I'll submit a new PR.

PR #324, opened 2016-12-02
No empty lines before '}' allowed

hasys:
This PR remove all empty lines before '}' symbol. For example:

```java
class SomeClass {

     public void emptyMethod() {

     }

    public void someMethod() {
         System.out.println("Hello World");

    }

}
```

will be replaced to:

```java
class SomeClass {

    public void emptyMethod() {
    }

    public void someMethod() {
        System.out.println("Hello World");
    }
}
```

To check this PR you can use IntelliJ editor:
File -> Settings -> Editor -> Code Style -> Java -> Blank Lines -> Keep Maximum Blank Lines -> Before '}':


psiroky:
And I like the following :)
```java
class SomeClass {

    public void emptyMethod() {}

    public void someMethod() {
        System.out.println("Hello World");
    }
}
```

ge0ffrey:
+1 for this change because the old Sun coding conventions (so the entire JDK) does it like this:
  http://www.oracle.com/technetwork/java/codeconventions-150003.pdf
My taste still prefers the other way, but following upstream conventions (sun, then jboss) is more important...

PR #331, opened 2016-12-12
[RHBRMS-2563] aligned org.jboss.resteasy to EAP 6.4.11

mbiarnes:


mbiarnes:
This PR has to be closed as the last public available version is 2.3.10.Final. 
2.3.15.Final is only existing as -redhat version.

PR #339, opened 2016-12-21
DROOLS-1374: Rename property version.wildfly.swarm to version.org.wil…

mbiarnes:
…dfly.swarm


csadilek:
@mbiarnes +1, but also needs to be changed in guvnor-ala before this can be merged:

https://github.com/droolsjbpm/guvnor/blob/master/guvnor-ala/guvnor-ala-distribution/pom.xml#L351

mbiarnes:
@csadilek : done! look at https://github.com/droolsjbpm/guvnor/pull/399

PR #342, opened 2017-01-03
Configure imports settings for stylecheck

hasys:
+ Fix for IDEA config file (disable star notation for *static* imports)
+ gitignore target folder not only in root but also for all sub directories (we need it for dashboard-builder-bom folder)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-/target
+target/
 /local
 
 # Eclipse, Netbeans and IntelliJ files
--- a/pom.xml
+++ b/pom.xml
@@ -372,6 +372,21 @@
                       <property name="severity" value="error"/>
                       <property name="eachLine" value="true"/>
                     </module>
+                    <module name="TreeWalker">
+                      <!-- Import settings: getting rid of redundant import, forbid star notation, defined order of imports. -->
+                      <module name="RedundantImport"/>
+                      <module name="AvoidStarImport">
+                        <property name="excludes" value="org.junit.Assert,org.assertj.core.api.Assertions,org.mockito.Mockito"/>
+                        <property name="allowClassImports" value="false"/>
+                        <property name="allowStaticMemberImports" value="false"/>
+                      </module>
+                      <module name="CustomImportOrder">
+                        <property name="customImportOrderRules"
+                                  value="STANDARD_JAVA_PACKAGE###THIRD_PARTY_PACKAGE###STATIC"/>
+                        <property name="sortImportsInGroupAlphabetically" value="true"/>
+                        <property name="separateLineBetweenGroups" value="true"/>
+                      </module>
+                    </module>
                   </module>
                 </checkstyleRules>
                 <outputFile>${project.build.directory}/checkstyle.log</outputFile>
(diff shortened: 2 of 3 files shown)


hasys:
Please, don't merge it before https://github.com/droolsjbpm/kie-wb-common/pull/621

.gitignore
```diff
@@ -1,4 +1,4 @@
-/target
+target/
```
ge0ffrey: -1 You should not ignore all target directories, only those with pom.xml
It's perfectly fine to call a java package target or to have a dataset for an example that has a directory called target.

pom.xml
```diff
@@ -372,6 +372,21 @@
...
+                        <property name="excludes" value="org.junit.Assert,org.assertj.core.api.Assertions,org.mockito.Mockito"/>
+                        <property name="allowClassImports" value="false"/>
+                        <property name="allowStaticMemberImports" value="false"/>
+                      </module>
+                      <module name="CustomImportOrder">
+                        <property name="customImportOrderRules"
+                                  value="STANDARD_JAVA_PACKAGE###THIRD_PARTY_PACKAGE###STATIC"/>
+                        <property name="sortImportsInGroupAlphabetically" value="true"/>
+                        <property name="separateLineBetweenGroups" value="true"/>
+                      </module>
```
ge0ffrey: Is this a pure subset of https://github.com/droolsjbpm/optaplanner/blob/master/pom.xml#L103 ?
So basically there is nothing in here that conflicts with the one that Ondrej proposed as checkstyle config for KIE in the long term (but which is too strict for most for now)?

.gitignore
hasys: Main goal was to ignore "dashboard-builder-bom/target". So is it OK if I will change it to:

```
/target
/dashboard-builder-bom/target
```

pom.xml
hasys: Hi @ge0ffrey ,
and yes, and no. Both of them have restrictions for Redundant Imports, proposed by Ondrej. Also Ondrej's config contains old java packages restrictions (like java.io and so on), but this PR not. On the other hand this rules contains Avoid Star Import module (with exceptions according to actual IDEA settings) and strict Imports Order (1) in addition.
(1.)
> **import** *java.*;
> **import** *javax.*; 
>
> **import** *other.packages*;
>
> **import static** *all.statics*;

So if you are using current IDEA settings it haven't to make any troubles to you. (Except this https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/342/files#diff-9d663bfb476dd8a86f4f213da123bfa5R4 fix)

Any way if your repository is affected you can use fixed IDEA settings for automated imports organizing:
Update your IDEA configuration and use Ctrl+Alt+O for single file or left mouse button on project name in Project Explorer -> Optimize Imports) for whole project, also you can set Optimize import on the fly option (Settings -> Editor -> General -> Auto Import -> Optimize import on the fly) and don't care how imports organized for new files :)

.gitignore
psiroky: Personally I think it would be better to have just single top-level `.gitignore` with `target/` and list the possible exceptions there (e.g. if we need to include some `target` dir). But that's just my view.

hasys: I agree with @psiroky, add exceptions will be better solution, then copy-paste `/target` from project to project for hundreds of time. As other benefit - if you have `/target` package - you will never forget to add exception, but it is so easy to forget to add gitignore with `target` to every project.

hasys:
We had to split this PR. See https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/356 and https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/357 for more info.

PR #344, opened 2017-01-03
upgraded errai version to 3.2.5.Final

mbiarnes:


manstis:
@mbiarnes I think you'll need similar for Uberfire too?

csadilek:
@manstis @mbiarnes working on the remaining updates with @mbarkley (there's also a sys property we need to enable for product.)

mbiarnes:
@csadilek can you pls confirm when you're done? I will re-start the kie-build then
@manstis uberfire and dashbuilder are done: https://github.com/uberfire/uberfire/pull/595 and https://github.com/dashbuilder/dashbuilder/pull/274

PR #346, opened 2017-01-05
Remove no longer needed kie-parent-with-dependencies

psiroky:
Child projects use directly kie-parent + import only the
BOMs they need.

In case there is a need to get all of the dependency versions
(internal + 3rd party ones), one can import the kie-platform-bom
which contains everything.

@mbiarnes, @ge0ffrey please let me know if agree with this. Thanks!


ge0ffrey:
+1

PR #355, opened 2017-01-12
PLANNER-707: Workbench "import example dialog": Embed stock example repositories

mcimbora:
Follow up after https://github.com/droolsjbpm/kie-wb-playground/pull/1

@ge0ffrey Can you please take a look?

### DIFF
--- a/kie-bom/pom.xml
+++ b/kie-bom/pom.xml
@@ -385,6 +385,13 @@
         <classifier>sources</classifier>
       </dependency>
 
+      <!-- KIE Workbench Playground Repository -->
+      <dependency>
+        <groupId>org.kie.workbench.playground</groupId>
+        <artifactId>playground-parent</artifactId>
+        <version>${version.org.kie}</version>
+      </dependency>
+
       <!--KIE Common - Core Services -->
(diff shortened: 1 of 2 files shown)


kie-bom/pom.xml
```diff
@@ -385,6 +385,19 @@
...
+      <dependency>
+        <groupId>org.kie.workbench.playground</groupId>
+        <artifactId>playground-parent</artifactId>
+        <version>${version.org.kie}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.kie.workbench.playground</groupId>
+        <artifactId>playground-parent</artifactId>
+        <version>${version.org.kie}</version>
+        <classifier>sources</classifier>
```
psiroky: I believe this is not needed, because the `playground-parent` uses packaging `pom` which does not produce binary jar nor sources jar.

psiroky:
Minor comment about the `playground-parent` deps, otherwise +1.

kie-bom/pom.xml
mcimbora: @psiroky I see, a good point.

PR #357, opened 2017-01-12
Stylecheck settings for Imports

hasys:
Part of https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/342 about CheckStyle configuration. We had to split it because we need to use IDEA style plugin to reformat Stunner before CheckStyle will be configured.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -354,6 +354,21 @@
                       <property name="severity" value="error"/>
                       <property name="eachLine" value="true"/>
                     </module>
+                    <module name="TreeWalker">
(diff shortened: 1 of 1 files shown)


hasys:
See also https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/356

hasys:
Hi @psiroky, hi @ge0ffrey.

Stunner is ready to go with it. All imports reformatted and now Stunner is able to be compiled with this restrictions. It can be merged, thank you for patient :)

PR #358, opened 2017-01-12
Wrap and align implemented interfaces, method declaration parameters and method call parameters

manstis:
Following the discussion with @hasys on https://github.com/droolsjbpm/kie-wb-common/pull/634 I fetched the latest IDEA code-formatter. Much to my surprise it generated quite different code format to (a) I am used to, (b) that appears to be used by the vast majority of other developers.

This PR therefore reinstates spaces 'within' constructs that appeared to be the norm in older versions of this code formatter and most certainly older versions of IDEA. Possibly affected by dc77c93f9d60507382831612f6ddf0024abba02a

### DIFF
--- a/ide-configuration/intellij-configuration/code-style/intellij-code-style_droolsjbpm-java-conventions.xml
+++ b/ide-configuration/intellij-configuration/code-style/intellij-code-style_droolsjbpm-java-conventions.xml
@@ -1,5 +1,5 @@
 <code_scheme name="KIE Java Conventions">
-  <option name="LINE_SEPARATOR" value="&#xA;" />
+  <option name="LINE_SEPARATOR" value="&#10;" />
   <option name="CLASS_COUNT_TO_USE_IMPORT_ON_DEMAND" value="99999" />
   <option name="NAMES_COUNT_TO_USE_IMPORT_ON_DEMAND" value="99999" />
   <option name="PACKAGES_TO_USE_IMPORT_ON_DEMAND">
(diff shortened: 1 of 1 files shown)


hasys:
Hi @manstis,

it is really strange situation. I tried version from Mar 4, 2011 (https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/blob/d93d2dbf2ff7166661e32db34c75392158abbaf9/ide-configuration/intellij-configuration/code-style/intellij-code-style_droolsjbpm-java-conventions.xml) in IDEA 11.1.5 and it handl wraps as it is in the current version (remove all whitspaces between parentheses and method parameters). So my guess is this style wasn't used or used some custom edition of this style.

ge0ffrey:
Ah, the joy of agreeing on a common code style, almost as much fun as determining which is the better religion (potentially with the Spanish inquisition, etc in the room). Anyway, I'd argue:

1) As far as I know, "spaces within constructs" isn't used by the vast majority of developers (not even just KIE developers). Outside of KIE, I never see it. Inside of KIE, I'd say is 50-50 (often within the same file!).
Even wb generated code has it for method params but not for annotations params (which is wierd).
2) The Sun Coding conventions don't have "spaces within constructors"
3) The JBoss coding conventions don't have "spaces withing constructors".
  https://github.com/jboss/ide-config/

Due to 2) and 3), most of our users find the coding style of our examples uncomfortable (because all other jboss and apache proejcts do it differently). I suggest we drop "spaces within constructs".

jhrcek:
+1 on dropping "spaces within constructs" and getting closer to the mainstream conventions!

manstis:
@jhrcek @ge0ffrey @hasys Not being one to rock the boat I've amended this PR to only wrap and align method declaration parameters, method call arguments and implemented interface declarations. This seems in keeping with http://www.oracle.com/technetwork/java/codeconventions-150003.pdf

ge0ffrey:
+1 to merge that "wrapping and aligning method declarations" to be in sync with java code conventions and jboss etc.

hasys:
Thank you @manstis +1

PR #371, opened 2017-01-26
Execute ban-duplicated-classes in full profile

psiroky:
This enforcer execution takes considerable amount of time,
especially for big modules like showcases and webapps which
further slows down the webapp start. Full profile will be
executed on Jekins to make sure we catch all the duplicated
classes before the changes get merged.


romartin:
Thanks @psiroky !!

PR #372, opened 2017-01-27
RHBPMS-4602: LinkageError during invocation of webservice workitem on WAS

manstis:
See https://issues.jboss.org/browse/RHBPMS-4602

See the JIRA comments for more details, but in summary:

```kie-parent``` ```pom.xml``` overrides ```xmlbeans``` version in ```jboss-integration-platform-bom``` from 2.6.0 to 2.3.0 - possibly for very good reason (that I am unaware). In order to use Apache POI with Microsoft Excel XLSX files we need to include ```org.apache.xmlbeans.XmlOptions``` that is not in IBM's JRE ```xml.jar```. However including ```xmlbeans-2.3.0.jar``` leads to duplicate ```org.w3c.dom.*``` classes and class loading problems.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -84,7 +84,6 @@
     <version.net.sf.docbook.docbook-xsl>1.76.1</version.net.sf.docbook.docbook-xsl>
 
     <version.org.apache.xmlgraphics.commons>1.4</version.org.apache.xmlgraphics.commons>
-    <version.org.apache.xmlbeans>2.3.0</version.org.apache.xmlbeans>
     <version.org.codehaus.cargo>1.6.2</version.org.codehaus.cargo>
     <version.org.jboss.arquillian.extension.drone>2.0.0.Final</version.org.jboss.arquillian.extension.drone>
     <version.org.jboss.arquillian.graphene>2.1.0.CR1</version.org.jboss.arquillian.graphene>
@@ -1149,6 +1148,15 @@
                         <ignoreClass>org.xmlpull.v1.XmlPullParser</ignoreClass>
                         <!-- Bundled by both com.sun:tools and com.sun.xml.bind:jaxb-xjc. No easy way to exclude one of them. -->
                         <ignoreClass>org.relaxng.datatype.*</ignoreClass>
+                        <!-- Classes duplicated by the jar itself (bug), see https://issues.apache.org.jira/browse/XMLBEANS-499 -->
+                        <ignoreClass>org.apache.xmlbeans.xml.stream.XMLName</ignoreClass>
+                        <ignoreClass>org.apache.xmlbeans.xml.stream.XMLInputStream</ignoreClass>
+                        <ignoreClass>org.apache.xmlbeans.xml.stream.utils.NestedThrowable</ignoreClass>
+                        <ignoreClass>org.apache.xmlbeans.xml.stream.utils.NestedThrowable$Util</ignoreClass>
+                        <ignoreClass>org.apache.xmlbeans.xml.stream.XMLStreamException</ignoreClass>
+                        <ignoreClass>org.apache.xmlbeans.xml.stream.ReferenceResolver</ignoreClass>
+                        <ignoreClass>org.apache.xmlbeans.xml.stream.XMLEvent</ignoreClass>
+                        <ignoreClass>org.apache.xmlbeans.xml.stream.Location</ignoreClass>
                       </ignoreClasses>
                       <findAllDuplicates>true</findAllDuplicates>
                     </banDuplicateClasses>

### CODE REVIEW

**Comment by manstis** [comment]
@psiroky @mswiderski If you know of any good reason why we were sticking to 2.3.0 please say.

@psiroky How can we trigger a full downstream build to get some confidence this does't break things?

**Comment by psiroky** [comment]

**Comment by psiroky** [comment]
There is no comment saying why the override is there, so that's bad. I have been able to track the change to https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/commit/1b7016dad7d58c22dc536d7fb12c5d32bdf432fd (3 years old commit, which was actually upgrading the xmlbeans). So IMO, the override needs to go and we need to fix whatever fails we encounter.

**Comment by mswiderski** [comment]
@manstis not aware of any reason why it should be 2.3.0

**Comment by manstis** [comment]

**Comment by manstis** [comment]
@psiroky Is this enabled now: 
```
15:31:35 [ERROR] Failed to execute goal org.apache.maven.plugins:maven-enforcer-plugin:1.4.1:enforce (ban-duplicated-classes) on project drools-decisiontables: Some Enforcer rules have failed. Look above for specific messages explaining why the rule failed. -> [Help 1]
15:31:35 org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-enforcer-plugin:1.4.1:enforce (ban-duplicated-classes) on project drools-decisiontables: Some Enforcer rules have failed. Look above for specific messages explaining why the rule failed.
```
I guess I'm expected to reformat the entire drools jbpm codebase to get this PR to compile. Sweet.

**Comment by psiroky** [comment]
@manstis If I understand the output correctly, the enforcer rule `ban-duplicated-classes` failed. This is not the checkstyle.

Looking thought the output I see
```
[WARNING] Rule 0: org.apache.maven.plugins.enforcer.BanDuplicateClasses failed with message:
16:28:26 Duplicate classes found:
16:28:26 
16:28:26   Found in:
16:28:26     org.apache.xmlbeans:xmlbeans:jar:2.6.0:compile
16:28:26   Duplicate classes:
16:28:26     org/apache/xmlbeans/xml/stream/XMLName.class
16:28:26     org/apache/xmlbeans/xml/stream/XMLEvent.class
16:28:26     org/apache/xmlbeans/xml/stream/utils/NestedThrowable.class
16:28:26     org/apache/xmlbeans/xml/stream/XMLInputStream.class
16:28:26     org/apache/xmlbeans/xml/stream/XMLStreamException.class
16:28:26     org/apache/xmlbeans/xml/stream/utils/NestedThrowable$Util.class
16:28:26     org/apache/xmlbeans/xml/stream/ReferenceResolver.class
16:28:26     org/apache/xmlbeans/xml/stream/Location.class
```
It's caused by faulty xmlbeans-2.6.0.jar, see https://issues.apache.org.jira/browse/XMLBEANS-499.

We need to workaround that by adding few more ignored classes for that enforcer rule:
```
<!-- Classes duplicated by the jar itself (bug), see https://issues.apache.org.jira/browse/XMLBEANS-499 -->
<ignoreClass>org.apache.xmlbeans.xml.stream.XMLName</ignoreClass>
<ignoreClass>org.apache.xmlbeans.xml.stream.XMLInputStream</ignoreClass>
<ignoreClass>org.apache.xmlbeans.xml.stream.utils.NestedThrowable</ignoreClass>
<ignoreClass>org.apache.xmlbeans.xml.stream.utils.NestedThrowable$Util</ignoreClass>
<ignoreClass>org.apache.xmlbeans.xml.stream.XMLStreamException</ignoreClass>
<ignoreClass>org.apache.xmlbeans.xml.stream.ReferenceResolver</ignoreClass>
<ignoreClass>org.apache.xmlbeans.xml.stream.XMLEvent</ignoreClass>
<ignoreClass>org.apache.xmlbeans.xml.stream.Location</ignoreClass>
```

manstis:
Thanks @psiroky I admit I didn't look in detail; but that is exactly the issue I was expecting (I'd read that ```xmlbeans-2.6.0.jar``` included duplicate class definitions). I don't think masking the duplicates at compile time is however the solution and fear the more sensitive Application Servers (see, I didn't explicitly say WAS!) may bork at runtime when loading classes. The internet is filled with people saying to remove the duplicates and re-package. Do you know who within RH would know if RH has a ```-redhat``` version with the duplicates removed? I realise this would only be a solution for our productized builds.

psiroky:
@manstis I guess the only way to know is to try. But AFAICT we only fool the enforcer, JDK/JRE should be able to handle this properly and load on one class (and it does not really matter which as they are in fact identical).

I do agree this needs to be properly fixed though.

Looking at https://maven.repository.redhat.com/ga/org/apache/xmlbeans/xmlbeans/ there is no `-redhat` rebuild of 2.6.0. What we can do is to release something like `2.6.0.jbossorg-1` where we would fix this. But that will of course require certain time to do.

manstis:

psiroky:
I gave this a brief look (sources here https://github.com/apache/xmlbeans) and the project is not Mavenized and can not even be currently built using Ant (broken links to 3rd party deps, etc). We would need to ideally Mavenize the project which may easily take days (and then we would need to promote this to upstream).

psiroky:
Or, of course, we just just re-package the jar, but that's not the best practice :)

manstis:

manstis:

manstis:
@psiroky WDYT? Merge?

psiroky:
@manstis honestly, I don't know. Having the build green is definitely great. I guess we could merge this, since we still have time to catch some side effects before the release?

manstis:
@psiroky OK, let's do it! Dare you approve? I don't want to raise, approve and merge my own PR.

PR #373, opened 2017-01-27
[6.5.x] RHBPMS-4602: LinkageError during invocation of webservice workitem on WAS

manstis:
See https://issues.jboss.org/browse/RHBPMS-4602

This is the corollary of https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/372 for 6.5.x


manstis:

manstis:
@psiroky WDYT? Dare we merge?

psiroky:
@manstis I have to say I wasn't expecting green build :) However, we don't test on WAS at all, so not sure what happens there and of course of tests don't cover everything. I would still leave this for the next patch.

manstis:
**DO NOT MERGE**

It is likely this PR will introduce duplicate class definitions that'll cause issues at runtime. 

See https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/372

PR #374, opened 2017-01-27
Next portion of checkStyle modules

hasys:
This PR enables following modules:

**NeedBraces**
Checks for braces around code blocks.
if, else, for, while, do...while.

**EqualsHashCode**
Checks that classes that either override equals() or hashCode() also overrides the other. This checks only verifies that the method declarations match Object.equals(Object)} and Object.hashCode() exactly to be considered an override. This check does not verify invalid method names, parameters other than Object, or anything else.

**UnusedImports**
Checks for unused import statements. Checkstyle uses a simple but very reliable algorithm to report on unused import statements. An import statement is considered unused if:
- It is not referenced in the file. The algorithm does not support wild-card imports like import java.io.*;. Most IDE's provide very sophisticated checks for imports that handle wild-card imports.
- It is a duplicate of another import. This is when a class is imported more than once.
- The class imported is from the java.lang package. For example importing java.lang.String.
- The class imported is from the same package.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -378,6 +378,11 @@
                         <property name="sortImportsInGroupAlphabetically" value="true"/>
                         <property name="separateLineBetweenGroups" value="true"/>
                       </module>
+
+                      <module name="NeedBraces">
+                        <property name="tokens" value="LITERAL_IF, LITERAL_ELSE, LITERAL_FOR, LITERAL_WHILE, LITERAL_DO"/>
+                      </module>
+                      <module name="EqualsHashCode"/>
                     </module>
                   </module>
                 </checkstyleRules>
@@ -1295,7 +1300,7 @@
           </exclusion>
         </exclusions>
       </dependency>
-      
+
       <!-- TODO BZ-996210 use org.apache.xmlgraphics:batik* instead -->
       <dependency>
         <groupId>batik</groupId>


hasys:
Stunner is ready to merge without any additional work.

psiroky:

ge0ffrey:
@psiroky The full downstream build won't validate if optaplanner is ok with this, I presume due to the override there?

+1 on NeedBraces and EqualsHashCode

As for UnusedImports - good idea if we can educate everyone to enable "remove unused imports on commit". Otherwise it's just a pita.

psiroky:
@ge0ffrey correct, optaplanner will override the settings and thus pass. I've triggered the build to make avoid the same situation as last time where some repos weren't ready and I had to do the override.

hasys:
Hi @ge0ffrey,

thank you for the review. I am agree with you, I guess UnusedImports will be a pain just for the first time for person. Also I can send an email to all developers with some information how to configure their IDE. Or we can remove UnusedImports for now and enable it later after all emails and discussions will be finished. What do you think? Is one time pain (and following IDE configuration) is not so bad, or it's better to postpone this module?

ge0ffrey:
@hasys +1 to merge if and only if such a mail is send to bsig-all. It's still a pain for contributors though, but they'll learn the hard way.

hasys:
@ge0ffrey, OK, thank you. I will send one with full explanation (I guess it will be at Monday).

hasys:
Jenkins retest please

hasys:

hasys:
**UnusedImports** removed from this PR because as I found Eclipse doesn't support some rules for imports as IDEA, so we need to recheck it. I will do it tomorrow (on Monday).

hasys:

hasys:

hasys:

hasys:

hasys:

hasys:
Hi @psiroky, this PR is ready to merge.

PR #378, opened 2017-02-01
Few more exceptions for ban-duplicate-class rule + exclude geronimo from Hibernate jars

psiroky:


psiroky:

mbiarnes:
+1

PR #381, opened 2017-02-07
enable code-coverage for GwtMockito tests

jsoltes:
By default jacoco java agent instruments classes on-the-fly during the test execution. This simplifies code coverage analysis a lot, as no pre-instrumentation and classpath tweaking is required.
However when using tests with custom classloader (tests that run with GwtMockitoTestRunner.class) the jacoco plugin shows zero coverage. This is because the ids of the classes in the report (jacoco.exec)
do not match the ids of the actual classes in the project (as GwtMockito uses proxies of the classes that are being tested). This problem has been solved by setting jacoco plugin to instrument
the classes before the test are run (and the proxies are created). This way exection report info is created from the original classes instead of the created proxies. The result is that the coverage of both
GwtMockito and "normal" tests can be measured. No negative side-effects (slowdown or misleading info in the report) have been observed.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -589,13 +589,16 @@
           <version>${version.jacoco.plugin}</version>
           <executions>
             <execution>
-              <id>default-prepare-agent</id>
+              <id>default-instrument</id>
               <goals>
-                <goal>prepare-agent</goal>
+                <goal>instrument</goal>
               </goals>
-              <configuration>
-                <propertyName>surefire.argLine</propertyName>
-              </configuration>
+            </execution>
+            <execution>
+              <id>default-restore-instrumented-classes</id>
+              <goals>
+                <goal>restore-instrumented-classes</goal>
+              </goals>
             </execution>
             <execution>
               <id>default-report</id>
@@ -1152,6 +1155,14 @@
           <name>code-coverage</name>
         </property>
       </activation>
+      <dependencies>
+        <dependency>
+            <groupId>org.jacoco</groupId>
+            <artifactId>org.jacoco.agent</artifactId>
+            <classifier>runtime</classifier>
+            <scope>test</scope>
+        </dependency>
+      </dependencies>
       <build>
         <pluginManagement>
           <plugins>
@@ -1159,8 +1170,9 @@
               <groupId>org.apache.maven.plugins</groupId>
               <artifactId>maven-surefire-plugin</artifactId>
               <configuration>
-                <!-- Append surefire.argLine property populated by JaCoCo plugin -->
-                <argLine>-Xmx1024m -Dfile.encoding=UTF-8 @{surefire.argLine}</argLine>
+                <systemPropertyVariables>
+                  <jacoco-agent.destfile>target/jacoco.exec</jacoco-agent.destfile>
+                </systemPropertyVariables>
               </configuration>
             </plugin>
           </plugins>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -1152,15 +1155,24 @@
...
+      </dependencies>
       <build>
         <pluginManagement>
           <plugins>
             <plugin>
               <groupId>org.apache.maven.plugins</groupId>
               <artifactId>maven-surefire-plugin</artifactId>
               <configuration>
-                <!-- Append surefire.argLine property populated by JaCoCo plugin -->
-                <argLine>-Xmx1024m -Dfile.encoding=UTF-8 @{surefire.argLine}</argLine>
```
triceo: I remember this being pretty important for JaCoCo. You sure it can be removed?

jsoltes: This property is used only by the prepare-agent goal. As this goal is replaced with instrument and restore-instrumented-classes goals, it is not needed anymore. This makes the java agent a little bit less flexible, but it is price that needs to be payed for the offline instrumentation. In the unlikely case you want to do some custom configurations to the java agent, there are two other ways how to do this (see http://www.eclemma.org/jacoco/trunk/doc/offline.html).

triceo: OK.

jhrcek:
@jsoltes Could you please add some basic info about how to enable code coverage analysis and where to look for the reports into the README?

jsoltes:
@jhrcek Thanks for the comment. I have added the short description of the usage into the README file.

manstis:
@jomarko @jsoltes @triceo IMO could QE please confirm this is good to merge?

I know little about the plugin and would simply be looking at formatting ;-)

triceo:
@manstis +1 in principle. I will leave it to @jomarko  to confirm that it works as expected.

manstis:
@jomarko Are you happy for this to be merged (along with https://github.com/droolsjbpm/kie-wb-common/pull/664)?

jomarko:
@manstis working on this. I found some projects that have failing tests after using new code-coverage profile. I am collecting them.

jomarko:
@manstis @jsoltes I found these [1] failing tests that fails after update of the code-coverage profile. In other projects works the profile well and measures coverage for GwtMocks correctly. I think this PR is worth to merge and the fails below can be solved in separate jira either with new code-coverage profile configuration or with updating of the tests.

[1]
* drools-wb-guided-dtable-editor-client
** ActionInsertFactPopupTest
** ConditionPopupTest
** ConditionPopupViewTest
* drools-wb-guided-rule-editor-client
** AttributeSelectorPopupTest

manstis:
@jsoltes Personally I think disabling existing tests because of a code coverage plugin is worse than not having the code coverage plugin running - as "gate keepers" we should not merge PRs without supporting unit tests; so IMO coverage reduction is unlikely (but not impossible) to reduce significantly in the interim between now and PRs being submitted for code coverage AND fixed tests. i.e. I'll merge this when PRs for the failing tests are also made.

psiroky:
Is anyone planning to look at those failing tests and possible fix them?

jhrcek:
@psiroky We're tracking this in our backlog. @jsoltes or @jomarko will prioritize this.

jsoltes:
@manstis I have made the pull-request with fix for the failing tests, see https://github.com/droolsjbpm/drools-wb/pull/417.

PR #387, opened 2017-02-16
Declare com.coderplus.maven.plugins:copy-rename-maven-plugin version information

manstis:
Add plugin information for ```com.coderplus.maven.plugins:copy-rename-maven-plugin```


manstis:
See https://github.com/droolsjbpm/kie-wb-distributions/pull/464 too

PR #388, opened 2017-02-16
kie-dmn needs to be built before jbpm

psiroky:
jBPM now depends on KIE DMN, so order needs to change


mswiderski:
+1, thank you @psiroky

PR #393, opened 2017-02-21
Introduce profile for revapi report generation

MarianMacik:
Hi, @psiroky,

here is the change we discussed recently. It allows to generate full revapi report including the ignored changes. See all related commits below.

Thanks

Marian

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -191,6 +191,8 @@
 
     <!-- Latest release to be used by api-compatibility-check to check backwards compatibility of the Kie API. -->
     <version.org.kie.latestFinal.release>6.5.0.Final</version.org.kie.latestFinal.release>
+    <revapi.oldKieVersion>${version.org.kie.latestFinal.release}</revapi.oldKieVersion>
+    <revapi.newKieVersion>${project.version}</revapi.newKieVersion>
 
     <!-- using undertow for kie server router - same version as it comes with WildFly 10.0.0.Final-->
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -191,6 +191,8 @@
 
     <!-- Latest release to be used by api-compatibility-check to check backwards compatibility of the Kie API. -->
     <version.org.kie.latestFinal.release>6.5.0.Final</version.org.kie.latestFinal.release>
+    <revapi.oldVersion>${version.org.kie.latestFinal.release}</revapi.oldVersion>
+    <revapi.newVersion>${project.version}</revapi.newVersion>
```
psiroky: I would suggest using more descriptive names, like `<revapi.kie.newVersion>` or maybe `<revapi.newKieVersion>`. I believe the `kie` is important to make it clear that the property expects a project (KIE) version.

pom.xml
```diff
@@ -476,12 +478,19 @@
...
+            <newArtifacts>
+              <artifact>${project.groupId}:${project.artifactId}:${revapi.newVersion}</artifact>
+            </newArtifacts>
             <!-- By default revapi will check the oldArtifact against the currently executed build -->
             <analysisConfigurationFiles>
               <configurationFile>
                 <path>src/build/revapi-config.json</path>
+                <roots>
+                  <root>filter</root>
+                  <root>ignores</root>
```
psiroky: Nit-picking - is there a reason for singular `filter` and plural `ignores`?

PR #396, opened 2017-02-23
JBPM-5714 - Add case management showcase to dependency version manage…

cristianonicolai:
…ment

### DIFF
--- a/jbpm-bom/pom.xml
+++ b/jbpm-bom/pom.xml
@@ -815,6 +815,12 @@
         <version>${version.org.kie}</version>
         <classifier>sources</classifier>
       </dependency>
+      <dependency>
+        <groupId>org.jbpm</groupId>
+        <artifactId>jbpm-wb-case-mgmt-showcase</artifactId>
+        <version>${version.org.kie}</version>
+        <type>war</type>
+      </dependency>
 
       <!-- designer and jbpm-simulation -->
       <dependency>


jbpm-bom/pom.xml
```diff
@@ -815,6 +815,12 @@
         <version>${version.org.kie}</version>
         <classifier>sources</classifier>
       </dependency>
+      <dependency>
+        <groupId>org.jbpm</groupId>
+        <artifactId>jbpm-wb-case-mgmt-showcase</artifactId>
+        <version>${version.org.kie}</version>
+        <type>war</type>
+      </dependency>
```
psiroky: Should we be pushing this unsupported showcase into the official jbpm-bom?

cristianonicolai: @psiroky this showcase will be used for demonstrating the case management runtime capabilities. Although is not part of the workbench, it will be provisioned as a separated and optional war.

psiroky: @cristianonicolai thanks for the details. What exactly does it mean? Should we publishing it on the jbpm.org pages? Or will it just be an link to the Nexus? We will do additional testing for that war?

cristianonicolai: @psiroky The showcase app is a demo that shows all the UI building blocks related to case management. The idea is that the app itself isn't supported but the individual blocks are.  The idea is that these can be consumed by custom apps.
As for publishing maybe @krisv can comment on that, I believe it could be just a link to nexus.
Regarding tests QA is already looking into that.

krisv: This case mgmt interface shouldn't really be treated differently from that point of view, it should be tested and will be delivered as part of the workbench.  Technically however, it is not hardcoded in the workbench but delivered as a separate war that can be provisioned into the workbench.

PR #397, opened 2017-02-24
Few more exceptions for duplicated classes rule

psiroky:


psiroky:

PR #407, opened 2017-03-10
Specifify the Maven Central before the JBoss.org repo to speed up builds

psiroky:
Testing the build time locally with empty local repo
(`mvn clean install -Dmaven.repo.local=maven-repo`) shows huge
improvement. 8min30s (without this change) vs 1min (with this change)
just for the build-bootstrap repo.

@mbiarnes, @ge0ffrey please take a look at let me know what you think. It definitely is not ideal, but I believe the speed up is worth it.

### DIFF
--- a/kie-user-bom-parent/pom.xml
+++ b/kie-user-bom-parent/pom.xml
@@ -58,10 +58,16 @@
   </properties>
 
   <repositories>
+    <!-- Having the <repositories> in this POM is an ugly hack to make sure we can always find
+         the jboss-integration-platform-parent (ip-parent). In most of the cases the ip-parent is available directly from
+         Maven Central. However, when new version gets released it takes up to 24 hours to get it synced into the Maven
+         Central. So in that period the local build of droolsjbpm-build-bootstrap would fail (unless one would
+         locally build the ip-parent from the sources as well, which is not something users should be doing). Configuring the
+         'repository.jboss.org' directly works around this issue. -->
     <repository>
       <!-- Duplicating the Maven Central repository here (as it is already coming from Super POM) makes the build much faster,
            as the Maven Central is now treated as the first (default) repository (because it is before the JBoss.org one).
-           Artifacts with release (fixed) versions are being downloaded primarily from there. Without the central being the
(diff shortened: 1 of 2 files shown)


psiroky:
@mbiarnes, @ge0ffrey any thoughts?

kie-user-bom-parent/pom.xml
```diff
@@ -74,6 +88,33 @@
       </snapshots>
     </repository>
   </repositories>
```
ge0ffrey: Why do we have repositories in kie-user-bom-parent's pom ?

In the root pom (kie-parent), it makes sense to have repositories, to bootstrap the build.
But in kie-user-bom-parent I would guess that if the user didn't already have the right repo, he wouldn't find the bom in the first place?

Note: also consider the productized version of this: does that have the maven customer repo instead?

ge0ffrey:
+1 to add it for the root pom.xml (kie-parent) if it doesn't cause havoc for our productized -redhat-7 versions (and I doubt it could as those need Maven Central too).

kie-user-bom-parent/pom.xml
psiroky: I would call this an ugly hack. When we release new IP BOM, it takes up to 24 hours to actually get published (synchronized) into Maven Central (yuck). But its available in repository.jboss.org right after it gets released. To avoid waiting before the upgrade, the repo is also here. I agree this is very far from ideal, but I am not aware of any better alternative right now.

ge0ffrey: If we keep this ugly hack, there needs to be a comment in the pom to explain why. We'll probably want to explicitly mention that this is make from sources builds succeed immediately after a release when Maven Central hasn't synced yet.

psiroky: I agree, I will add the clarification.

psiroky: I've the added comment https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/407/commits/55a6c9e78e560c45465121f644dd36c4a107e467

PR #412, opened 2017-03-16
improved scripts

mbiarnes:


psiroky:
@mbiarnes there is no need to create PRs if you are not expecting a review. Just push the changes directly.

PR #414, opened 2017-03-17
DSL scripts: improved scripts

mbiarnes:


psiroky:
@mbiarnes please don't create PRs if you don't expect review (and also don't expect green light from Jenkins). Just push the changes directly.

PR #417, opened 2017-03-22
Upgrade WildFly Swarm from 2016.9 to 2017.3.3

psiroky:


psiroky:

psiroky:

psiroky:
execute full downstream build

psiroky:

mswiderski:
@psiroky just fyi I had major issues with swarm 2017.3.3 and it turned out to be caused by this https://issues.jboss.org/browse/SWARM-1209

In general this makes kie server on swarm to not start at all due to corrupted configuration

psiroky:
@mswiderski thanks for the heads up. I'll see if the build works or not. 2017.4.0 should be out in a week or two, so we could also wait a bit longer.

psiroky:

psiroky:
@mswiderski are you OK with merging this if the build becomes green? We should definitely then upgrade to 2017.4.0 once it gets out.

mswiderski:
fine with me if that works for the bits we have relying on it. The kie server one is standalone component that I use as demo/playground so not affected by this one at the moment.

psiroky:
So I am giving up on trying to make this green. To best of my knowledge none of the randomly failing tests is related to these changes.

PR #418, opened 2017-03-22
Update com.lowagie:itext exclusions

psiroky:
@krisv, @porcelli I believe this should fix the issue (even though I can not reproduce it locally). I was looking at last Beta release and the WARs did not contain the bouncycastle deps, so it should be safe to exclude them.

This workaround can be removed once we upgrade to IP BOM 7.0.0.CR10+.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1647,6 +1647,28 @@
       </dependency>
 
       <dependency>
+        <!-- Can be removed once we upgrade to IP BOM 7.0.0.CR10+ -->
+        <groupId>com.lowagie</groupId>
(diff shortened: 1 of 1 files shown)


psiroky:

krisv:
+1

psiroky:

psiroky:
@livthomas FYI this will also fix the deployment issues on Tomcat.

psiroky:
So I am giving up on trying to make this green. To best of my knowledge none of the randomly failing tests is related to these changes.

PR #420, opened 2017-03-24
GUVNOR-2933: Guvnor ALA integration

wmedvede:


wmedvede:
Hello @porcelli, this PR is for the maven build integration we have been chatting about.
Would you mind review.

Thanks.

guvnor-bom/pom.xml
```diff
@@ -164,6 +164,12 @@
       </dependency>
       <dependency>
         <groupId>org.guvnor</groupId>
+        <artifactId>guvnor-ala-build-maven</artifactId>
+        <version>${version.org.kie}</version>
+        <classifier>tests</classifier>
```
porcelli: @wmedvede i don't think we use classifiers here

wmedvede: @porcelli I'll double check, but in the past, on other modules, I had to add the "tests" classifier in to be able to use e.g. some TestBase in guvnor, in other test class in kie-wb-commons. And this is basically the case here, I'm interested in using some test classes from guvnor in kie-wb-commons

psiroky: There are two options for test jars:
 1. `<classifier>tests</classifier>`
 2. `<scope>test</scope>`

We are mixing both in different POMs, so it probably does not matter which one gets used.

porcelli: @wmedvede k, got what you meant;

wmedvede: @psiroky I see, thanks. in this case the <classifier>tests</classifier> is being used.

PR #421, opened 2017-03-27
DSL-scripts: automatically update of latestReleasedVersionFromThisBra…

mbiarnes:
…nch and jbpm-installer

### DIFF
--- a/script/release/DSL-scripts/KIE_pushReleaseBranches.sh
+++ b/script/release/DSL-scripts/KIE_pushReleaseBranches.sh
@@ -47,13 +47,13 @@ fi
 
 # update kie-parent-metadata
 cd droolsjbpm-build-bootstrap/
-
 # change <version.org.uberfire>, <version.org.dashbuilder> and <version.org.jboss.errai>
 sed -i "$!N;s/<version.org.uberfire>.*.<\/version.org.uberfire>/<version.org.uberfire>$UBERFIRE_VERSION<\/version.org.uberfire>/;P;D" pom.xml
 sed -i "$!N;s/<version.org.dashbuilder>.*.<\/version.org.dashbuilder>/<version.org.dashbuilder>$DASHBUILDER_VERSION<\/version.org.dashbuilder>/;P;D" pom.xml
 sed -i "$!N;s/<version.org.jboss.errai>.*.<\/version.org.jboss.errai>/<version.org.jboss.errai>$ERRAI_VERSION<\/version.org.jboss.errai>/;P;D" pom.xml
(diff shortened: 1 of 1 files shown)


mbiarnes:
@psiroky could you please review this and ack it - so it can be merged and I can do a final test during the release tomorrow.
I received complaints when pushing directly to other projects, so I prefer to raise PRs and wait for the ack

script/release/DSL-scripts/KIE_pushReleaseBranches.sh
```diff
@@ -47,13 +47,16 @@ fi
...
 # change <version.org.uberfire>, <version.org.dashbuilder> and <version.org.jboss.errai>
 sed -i "$!N;s/<version.org.uberfire>.*.<\/version.org.uberfire>/<version.org.uberfire>$UBERFIRE_VERSION<\/version.org.uberfire>/;P;D" pom.xml
 sed -i "$!N;s/<version.org.dashbuilder>.*.<\/version.org.dashbuilder>/<version.org.dashbuilder>$DASHBUILDER_VERSION<\/version.org.dashbuilder>/;P;D" pom.xml
 sed -i "$!N;s/<version.org.jboss.errai>.*.<\/version.org.jboss.errai>/<version.org.jboss.errai>$ERRAI_VERSION<\/version.org.jboss.errai>/;P;D" pom.xml
+sed -i "$!N;s/<latestReleasedVersionFromThisBranch>.*.<\/latestReleasedVersionFromThisBranch>/<latestReleasedVersionFromThisBranch>$RELEASE_VERSION<\/latestReleasedVersionFromThisBranch>/;P;D" pom.xml
 
-cd ..
+cd ../jbpm
+# change release.version in jbpm/bpm-installer/src/main/resources/build.properties
+sed -i "$!N;s/release.version=.*./release.version=$RELEASE_VERSION/;P;D" jbpm-installer/src/main/resources/build.properties
```
psiroky: This should not be necessary. The maven module itself needs to update the version accordingly during the build. Or am I missing something here?

mbiarnes:
@psiroky during tests I saw that it was not upgraded after the version update. I saw there two SNAPSHOTs, one for the SNAPSHOT version and one for the RELEASE version. I will release today not using this change and review it again if it changed automatically.

psiroky:
@mbiarnes ok, thanks for info. I believe we should reduce this `sed`-style magic as much as possible (I know its not possible in all of the cases). This one seems like a good candidate for simple resource filtering.

mbiarnes:
@psiroky checked. It was changed to the right version. I removed it.

mbiarnes:
@psiroky I will close this PR since I see that also the property latestReleasedVersion is updated.
This PR is obsolete then.

PR #432, opened 2017-04-13
Use WildFly 10.1.0.Final for GWT SDM

psiroky:


manstis:
If/when this is merged can you please email BSIG as all developers IDE configuration will need changing too.

psiroky:
@manstis not sure if it makes sense to upgrade. I did it just because I was upgrading some other versions. If it means everyone has to mess with their configs, then it is perhaps better to just use the current version.

psiroky:
jenkins retest this

ge0ffrey:
@manstis @psiroky Can we merge this PR anyway?

psiroky:
@ge0ffrey I don't have anything against merging, just wasn't sure if it worth the trouble.

manstis:
Let's get some clear communication to BSIG first; as once merged everyone will need to update their IDE launch configurations to use WF 10.1.0.Final. When the proposed upgrade has been communicated and we've allowed a day or so for it to sink into peoples psyche we can merge.

psiroky:
E-mail sent to BSIG. Let's wait few more days and then merge.

ge0ffrey:
I think it can be merged now?

manstis:
I think so..

PR #438, opened 2017-05-10
added scripts for kie, uberfire and dashbuilder release

mbiarnes:


script/release/dash_createAndDeploy.sh
```diff
@@ -0,0 +1,52 @@
+# removing dashbuilder artifacts from local maven repo (basically all possible SNAPSHOTs)
```
psiroky: Do we want to have these dash and uf scripts here? Shouldn't they be in the dashbuilder and uberfire repos?

psiroky:
I would remove all the scripts which are just downloading the -project-sources.tar.gz. We should either have one single generic script which accepts parameters or move this directly to the repo with the DSL scripts. @mbiarnes WDYT?

mbiarnes:
@psiroky can we merge this and do slight changes in future?

psiroky:
@mbiarnes I am not really happy about this, but I guess we don't have any other option at this point. So go ahead and merge this.

mbiarnes:
@psiroky I promise that you will be happy one day ;)

PR #449, opened 2017-05-30
Added wildfly version property

Christopher-Chianelli:


Christopher-Chianelli:
Related/Required for https://github.com/kiegroup/kie-wb-common/pull/910

pom.xml
```diff
@@ -92,6 +92,10 @@
 
     <version.org.drools.guvnor-api>5.6.0.Final</version.org.drools.guvnor-api>
     <version.org.jboss.errai>4.0.1-SNAPSHOT</version.org.jboss.errai>
+
+    <!-- Required for kie-wb-common to build in IntelliJ Idea 2017 -->
+    <version.org.jboss.errai.wildfly>10.0.0.Final</version.org.jboss.errai.wildfly>
```
ge0ffrey: Should be "version.org.wildfly" instead and let's use the latest and greatest:

    <version.org.wildfly>10.1.0.Final</version.org.wildfly>

optaplanner-webexamples already has this, but it should be removed there once this PR is merged.

mbarkley: This property is meant for [this](https://github.com/errai/errai/tree/master/errai-cdi/errai-wildfly10-dist) special version of wildfly-dist packaged in Errai. This wildfly-dist removes files in the normal wildfly-dist to workaround [this bug](https://youtrack.jetbrains.com/issue/IDEA-166417) in IntelliJ.

ge0ffrey: If it's a temporary dependency, let's add a link in the comment in the xml, something like `<!-- TODO remove m when this is fixed: https://youtrack.jetbrains.com/issue/IDEA-166417 -->`

If it's not a temporary dependency, we should probably align the wildfly version through all of kie, to make sure we have at least one wildfly version on which everything works (so at least one version that a project that combines everything is sure to work on). It also makes it easier to try other versions (think QA certification etc). The downside is that if anyone upgrades that version, it might break something in other the kie projects, but that's the lesser evil in our experience... that can be fixed with asking jenkins for a full downstream build.

ge0ffrey:
ok to test

pom.xml
mbarkley: I imagine we'll want to remove this as soon as the IntelliJ bug is fixed, so I think the comment is a good idea.

Christopher-Chianelli:
Added.

porcelli:
@Christopher-Chianelli squash please?

Christopher-Chianelli:
Squashed

PR #450, opened 2017-06-05
Libraries used by the Maven incremental build

desmax74:
Dependencies used by the Maven Incremental Compiler in the kie-wb-common

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -206,10 +206,8 @@
     <version.org.jboss.forge.roaster>2.19.4.Final</version.org.jboss.forge.roaster>
 
     <!-- deps for incremental compilation -->
-    <version.org.apache.maven.incremental>3.3.9</version.org.apache.maven.incremental>
-    <version.plexus.interpolation>1.21</version.plexus.interpolation>
     <version.commons.cli>1.3.1</version.commons.cli>
-    <version.plexus.logging>1.0.3</version.plexus.logging>
+    <version.google.collections>1.0</version.google.collections>
   </properties>
 
   <repositories>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -204,6 +204,12 @@
     <!-- Overrides the version in the jboss-integration-platform-bom. A peer PR will be sent to jboss-integration-platform project.
          Please move this new version number there when possible -->
     <version.org.jboss.forge.roaster>2.19.4.Final</version.org.jboss.forge.roaster>
+
+    <!-- deps for incremental compilation -->
+    <version.org.apache.maven.incremental>3.3.9</version.org.apache.maven.incremental>
+    <version.plexus.interpolation>1.21</version.plexus.interpolation>
+    <version.commons.cli>1.3.1</version.commons.cli>
+    <version.plexus.logging>1.0.3</version.plexus.logging>
```
psiroky: These should already be coming from the ip-bom (https://github.com/jboss-integration/jboss-integration-platform-bom). The only one we will likely need to upgrade is the {{commons-cli}}.

desmax74:
Is it ok ?

pom.xml
```diff
@@ -2232,6 +2236,19 @@
         <version>${version.org.w3c.css.sac}</version>
       </dependency>
 
+      <!-- libraries for incremental compilation-->
+      <dependency>
+        <groupId>commons-cli</groupId>
+        <artifactId>commons-cli</artifactId>
+        <version>${version.commons.cli}</version>
+      </dependency>
```
psiroky: Sorry for nit-picking, but this is technically not needed. Overriding the version property is enough (as the kie-parent directly inherits from ip-bom).

pom.xml
```diff
@@ -2232,6 +2236,19 @@
...
+
+      <dependency>
+        <groupId>com.google.collections</groupId>
+        <artifactId>google-collections</artifactId>
+        <version>${version.google.collections}</version>
+      </dependency>
```
psiroky: Looks like the latest version of `google-collections` is almost 8 years old. Do we want to add this kind of dependency? Which module requires this?

desmax74: Maven internal API :(

desmax74:
PR outdated, it can be closed.

PR #453, opened 2017-06-09
UF-506: Added dependencies for concurrenci-api

adrielparedes:
Added dependencies for concurrency api. We need them to use ManagedExecutorService.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -214,6 +214,7 @@
     <!-- Overrides the version in the jboss-integration-platform-bom. A peer PR will be sent to jboss-integration-platform project.
          Please move this new version number there when possible -->
     <version.org.jboss.forge.roaster>2.19.4.Final</version.org.jboss.forge.roaster>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -214,6 +214,7 @@
     <!-- Overrides the version in the jboss-integration-platform-bom. A peer PR will be sent to jboss-integration-platform project.
          Please move this new version number there when possible -->
     <version.org.jboss.forge.roaster>2.19.4.Final</version.org.jboss.forge.roaster>
+    <version.org.jboss.concurrency-api>1.0.0.Final</version.org.jboss.concurrency-api>
```
porcelli: maybe use the same format as you use in uberfire: version.org.jboss.spec.javax.enterprise.concurrent? I don't know what is the correct way of doing this... I'll let @psiroky comment

psiroky: Yes, generally we should be using `version.<group-id>` where possible. If the property is ambiguous, `artifact-id` can be appended.

porcelli: @adrielparedes mind change here to version.org.jboss.spec.javax.enterprise.concurrent ?

PR #464, opened 2017-06-25
JBPM-4480: Add Swagger to KIE-Server REST API

DuncanDoyle:
This is for JBPM-4480: https://issues.jboss.org/browse/JBPM-4480

One additional comment, the POM already contained and old com.wordnik.swagger dependency, but I couldn't really find where that dependency is actually used. If it's not used anywhere, we should remove it. If it is used, we should consider upgrading the dependency to "io.swagger".

### DIFF
--- a/kie-bom/pom.xml
+++ b/kie-bom/pom.xml
@@ -225,6 +225,17 @@
       </dependency>
       <dependency>
         <groupId>org.kie.server</groupId>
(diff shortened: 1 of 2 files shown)


mswiderski:
jenkins retest this

mswiderski:
last retest to make sure it can be safely merged and of so will merge :)

PR #484, opened 2017-07-27
jgit upgrade needed by appformer - can be removed once moved to IP BOM

porcelli:
ps: this should fix the build

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -147,6 +147,10 @@
     <version.com.ahome-it.lienzo-core>2.0.287-RELEASE</version.com.ahome-it.lienzo-core>
     <version.com.ahome-it.lienzo-tests>2.0.287-RELEASE</version.com.ahome-it.lienzo-tests>
 
+    <!-- JGIT 4.8.0.201706111038-r required by UberFire/AppFormer. Can be removed once IP BOM is upgraded.-->
+    <version.org.eclipse.jgit>4.8.0.201706111038-r</version.org.eclipse.jgit>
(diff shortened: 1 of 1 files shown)


hasys:

psiroky:
@hasys just FYI: no need to ask for the downstream build explicitly. It will be executed automatically once the standard build succeeds.

pom.xml
```diff
@@ -147,6 +147,10 @@
     <version.com.ahome-it.lienzo-core>2.0.287-RELEASE</version.com.ahome-it.lienzo-core>
     <version.com.ahome-it.lienzo-tests>2.0.287-RELEASE</version.com.ahome-it.lienzo-tests>
 
+    <!-- JGIT 4.8.0.201706111038-r required by UberFire/AppFormer. Can be removed once IP BOM is upgraded.-->
+    <version.org.eclipse.jgit>4.8.0.201706111038-r</version.org.eclipse.jgit>
+    <version.org.apache.sshd>1.6.0</version.org.apache.sshd>
```
psiroky: I know it's a pain, but it would really be great to have a IP BOM PR as well.

porcelli: will IP bom be kept on v7?

psiroky: Do we want/need this change only for 7.x? If so, we should probably branch ip-bom as well.

porcelli: yes.. only for 8.x

porcelli:
the full build probably will fail, as kie-wb-common is failing;
and the fix for kie-wb-common (https://github.com/kiegroup/kie-wb-common/pull/996) is failing because is missing this issue

PR #488, opened 2017-07-28
Add branch mapping file

psiroky:
@mbiarnes please take a look. This should make the version upgrades easier. Instead of updating the 
`kie-build-helper-jenkins-plugin` every time, we would just need to update this file. One thing to remember is that we need to make sure the file is valid at all times otherwise bad things would happen :)

[jenkins skip ci]


psiroky:
@mbiarnes I haven't yet updated the plugin. My aim is to _not_ require any changes in the plugin when updating versions.

PR #514, opened 2017-08-16

mbiarnes:


pom.xml
```diff
@@ -120,6 +120,8 @@
     <version.org.springframework>4.3.2.RELEASE</version.org.springframework>
     <!-- Required to support SNI requests, see RHBRMS-2870 -->
     <version.org.apache.httpcomponents.httpclient>4.5.3</version.org.apache.httpcomponents.httpclient>
+    <!-- RHBRMS-2870 override of httpcomponents.httpcore -->
```
psiroky: Nit picking: the comment should ideally say why the override is needed, not that it is an override.

PR #518, opened 2017-08-17

mbiarnes:


mbiarnes:
this PR should be closed as it adds httpcore:4.4.6  that is the version in jboss-ip-bom 7.0.0.Final. Once merged the PR #528.

PR #519, opened 2017-08-17

mbiarnes:
(cherry picked from commit 1b3251f5db803460b55a00bab9f606106833ae03)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -101,6 +101,7 @@
 
     <!-- Required to support SNI requests, see RHBRMS-2870 -->
     <version.org.apache.httpcomponents.httpclient>4.5.3</version.org.apache.httpcomponents.httpclient>
+    <version.org.apache.httpcomponents.httpcore>4.4.6</version.org.apache.httpcomponents.httpcore>
 
     <version.org.drools.guvnor-api>5.6.0.Final</version.org.drools.guvnor-api>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -2426,10 +2427,25 @@
...
+        <groupId>org.apache.httpcomponents</groupId>
         <artifactId>httpmime</artifactId>
         <version>${version.org.apache.httpcomponents.httpclient}</version>
       </dependency>
       <dependency>
+        <groupId>org.apache.httpcomponents</groupId>
+        <artifactId>httpcore</artifactId>
+        <version>${version.org.apache.httpcomponents.httpcore}</version>
+      </dependency>
+      <dependency>
```
psiroky: Is this needed? I believe we only need to override the property?

mbiarnes: It seems if you only override the property but not explicitly the artifact the artifact coming from jboss-ip-bom has the version of ip-bom.
httpcore was not upgraded to the version we did an override in kie-parent

psiroky: This was the case for 6.x, but for 7.x the property override is enough. Just verified that locally with effective pom.

mbiarnes:
@psiroky done

psiroky:

psiroky:
Failure not related (and the downstream build already failed few times for different reasons). The same change is already on master and 6.5.x, so it should be pretty safe.

PR #532, opened 2017-08-24
Upgrade freemarker to 2.3.26.jbossorg-1

psiroky:
See https://github.com/AppFormer/uberfire/pull/813 for bit more info.

PR #538, opened 2017-08-30
Upgrade freemarker to 2.3.26.jbossorg-1

psiroky:
Backport of #532


psiroky:

psiroky:
The failures are not related and the full downstream build already failed 8 times in many different ways. Merging this as it already works on master and it will decrease the the number of false negatives.

PR #539, opened 2017-09-04
JBPM-6388 - Fill in wid entries based on annotations in workitem hand…

tsurdilo:
…ler (jbpm-workitems-archetype)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -87,6 +87,7 @@
     <version.com.github.detro>1.2.0</version.com.github.detro>
     <version.com.github.tomakehurst.wiremock>1.53</version.com.github.tomakehurst.wiremock>
     <version.com.google.android>4.1.1.4</version.com.google.android>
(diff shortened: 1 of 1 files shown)


tsurdilo:
@psiroky here is the pr for adding the depends to bootstrap. thanks.

pom.xml
```diff
@@ -1806,6 +1807,13 @@
       </dependency>
 
       <dependency>
+        <groupId>com.google.testing.compile</groupId>
+        <artifactId>compile-testing</artifactId>
+        <version>${version.com.google.testing.compile}</version>
+        <scope>test</scope>
```
psiroky: Nit picking -- we should not specify `scope` in the `depMgmt` section. This should be done in the module which directly uses the dependency.

tsurdilo:
@psiroky updated - thanks!

PR #543, opened 2017-09-07
Add repo `drlx-parser` to the list of repositories

tarilabs:
before `drools`

/cc @mariofusco

### DIFF
--- a/script/repository-list.txt
+++ b/script/repository-list.txt
@@ -2,6 +2,7 @@ kie-soup
 appformer
 droolsjbpm-build-bootstrap
 droolsjbpm-knowledge
+drlx-parser
 drools
 optaplanner
 jbpm


psiroky:
@ryanzhang @AntStephenson we will very likely need to disable maven deploy for some of the modules in the new repo https://github.com/kiegroup/drlx-parser. There is a need to build (but not deploy) some of moduels as they are then uberjared together. Only the uberjar is of the interest.

I remember skipping deploy is not an option in Brew. So we would end-up with the modules deployed anyway (for the product builds). Is that correct? Do you see any other solution for this?

tarilabs:
Adding some more details, in case needed.

repo kiegroup/drlx-parser is a fork of an upstream project (also available on github and nexus) called javaparser.

It is composed of:
```
[INFO] Reactor Build Order:
[INFO] 
[INFO] javaparser-parent
[INFO] javaparser-core
[INFO] javaparser-testing
[INFO] javaparser-core-generators
[INFO] javaparser-metamodel-generator
[INFO] drlx-parser
```

Please notice the last module `drlx-parser` is introduced by us, and it's an uber jar.

We locally use `mvn clean install` and works without a problem.

However, when on Jenkins, we would need only the module `drlx-parser` to be published.

This is because from drools we use only internally the module `drlx-parser` which is an uberjar of the whole forked project and our own extensions.
The module `drlx-parser` also shades the package names to avoid collisions; this allows end-users to have in their dependencies drools and upstream-javaparser without a problem.

AntStephenson:
It is true, Brew requires deployment to be true and skipping it would cause errors. It's also true that we override the deployment to always be true when building the product to avoid this error. 

I don't really see any other solution at the moment, is skipping deployment in the community but forcing deployment in product feasible? I'm also wondering, in addition to this, if we can exclude these artifacts from the offline maven repo, this would then not be added to MRRC (although I'm not entirely sure it's OK to do this).

psiroky:
@AntStephenson if we go with the deploy skip, then we must make sure the exclude the artifacts from the offline/online maven repos. They are just intermediate artifacts and we can't publish them in any way.

I will try another approach using the maven-invoker-plugin. If that won't work we will likely need to go with the deploy skip opiton.

AntStephenson:
@psiroky OK, please let me know what you go with in the end and I can update the exclusions before we forget :)

mbiarnes:
@psiroky @AntStephenson IIUC drlx-parser will be part of the repository-list.txt. But when this will be the only rep that will not be deployed - only built? Petr will try a solution with maven-invoker-plugin.
My question: is this rep also needed in gerrit? It is important to know if the tags that wil be  pushed to gerrit will include this repo or not? (we had a similar problem on gerrit with a missing rep [droolsjbpm-tools] that was not pushed. Scripts using the repository-list.txt, which has droolsjbpm-tools will fail - )

tarilabs:
rebased.

tarilabs:
To avoid a big potential issue as there is no confirmation of skipped deployment and for only the shaded uberjar to be the only one deployed an published, I will close this PR temporarily until we can clarify the process is fully aligned.

tarilabs:
Reopening as a missing prerequisite (kiegroup/drlx-parser#6) has now been fixed.

tarilabs:

tarilabs:
```
Test Result (1 failure / +1)
org.drools.workbench.screens.guided.scorecard.backend.server.indexing.IndexGuidedScoreCardTest.testIndexGuidedScoreCard
```

unrelated failure.

mbiarnes:

PR #547, opened 2017-09-08
Update IDEA code style not to wrap method params and call arguments

yurloc:
Always automatically wrapping method declaration parameters and call arguments is an inconvenient requirement. It [harms readability](https://github.com/jhrcek/guvnor/blob/a81b8781fca06bfdb99fb19a0ccc35314f206b13/guvnor-ala/guvnor-ala-registry-local/src/test/java/org/guvnor/ala/registry/tests/PageSortUtilsTest.java#L81-L101) and is a destructive operation as no IDE is able to revert it even if the rule is lifted. Better to fix the template as soon as possible before more [repositories](https://github.com/kiegroup/optaplanner-wb/pull/179) are [affected](https://github.com/kiegroup/guvnor/pull/478).

### DIFF
--- a/ide-configuration/intellij-configuration/code-style/intellij-code-style_droolsjbpm-java-conventions.xml
+++ b/ide-configuration/intellij-configuration/code-style/intellij-code-style_droolsjbpm-java-conventions.xml
@@ -1,5 +1,5 @@
 <code_scheme name="KIE Java Conventions">
-  <option name="LINE_SEPARATOR" value="&#10;" />
+  <option name="LINE_SEPARATOR" value="&#xA;" />
   <option name="CLASS_COUNT_TO_USE_IMPORT_ON_DEMAND" value="99999" />
   <option name="NAMES_COUNT_TO_USE_IMPORT_ON_DEMAND" value="99999" />
   <option name="PACKAGES_TO_USE_IMPORT_ON_DEMAND">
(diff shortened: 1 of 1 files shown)


jhrcek:
+1 for disabling this rule, but please let's gather wider consensus before  merging this. If this initiative is finally successfully, most people will need to update their locally imported formatting config files and start using it consistently. Also we could then finally execute our planned "format all repos using single agreed-upon code style". Any ideas how to engage more people to approve this?

ide-configuration/intellij-configuration/code-style/intellij-code-style_droolsjbpm-java-conventions.xml
```diff
@@ -73,8 +73,6 @@
     <option name="ALIGN_MULTILINE_PARAMETERS_IN_CALLS" value="true" />
     <option name="ALIGN_MULTILINE_FOR" value="false" />
     <option name="ALIGN_MULTILINE_EXTENDS_LIST" value="true" />
-    <option name="CALL_PARAMETERS_WRAP" value="2" />
-    <option name="METHOD_PARAMETERS_WRAP" value="2" />
```
yurloc: @hasys Can you check if eclipse rules are currently in sync with this?

hasys: It is not 100% similar, because it is not possible, but it is similar now. Main points are:
* Eclipse won't override alignment from IDEA in this place
* you can man manually wrap parameters on your choose and Eclipse won't change it back
* parameters will be aligned in a column if new line present

yurloc: Thanks Kirill, that actually sounds like 100% IDEA's behavior. Time will tell...

PR #548, opened 2017-09-13
upgrade mvel to version 2.3.2.Final

mariofusco:


mariofusco:

pom.xml
```diff
@@ -2525,6 +2527,11 @@
         <artifactId>undertow-websockets-jsr</artifactId>
         <version>${version.io.undertow.core}</version>
       </dependency>
+      <dependency>
+        <groupId>org.mvel</groupId>
+        <artifactId>mvel2</artifactId>
+        <version>${version.org.mvel}</version>
+      </dependency>
```
psiroky: This is not necessary as we inherit the depMgmg from the ip-bom directly, but it won't hurt anything, so no need to block the merge.

PR #551, opened 2017-09-14
added scripts for automated creation of jBPM-installer.zip and jBPM-i…

mbiarnes:
…nstaller-full.zip


psiroky:
@mbiarnes looks nice! Would it be possible to split the create and push tasks? In the current form, it is very hard to test this locally. I think we should generally have 1 or more "create" scripts and then 1 push script, which takes everything and pushes it via scp to the data servers. WDYT?

mbiarnes:
sure - there are changes raising tight now to sync with prod (json file,
gerrit).
Once this is working I will refactor the whole reelase process improving
the DSL jobs. But for theis I want a meetg. with you before.

mbiarnes:
@psiroky done - jbpm-installer and jbpm-installer-full.zip in one script
Now I will fix the related kie-jenkins-script PR and then backport this to 7.4.x

psiroky:
Does my first comment make sense? I still think it would be better to separate the concerns and have two different scripts - one to create all the zips + second one which push it to wherever we need it to. WDYT?

mbiarnes:
OK -

mbiarnes:
@psiroky like this?

PR #559, opened 2017-09-25
jbpm-case-mgmt is not exist anymore, so remove it from jbpm-bom

ryanzhang:


ryanzhang:
@mbiarnes Hi, Michael 
it seems that the build can't fetch some artifacts from nexus which seems irrelevant from the PR.
Could you please have a look if the build works OK.

ryanzhang:
@psiroky Petr It seems that the build error is caused by can't download org.freemarker:freemarker:jar:2.3.26.jbossorg-1 ?

psiroky:
@ryanzhang that was just a glitch/error in the nexus I believe. The PR is green now.

PR #568, opened 2017-10-09
changed for 7.4.x branch

mbiarnes:


script/branch-mapping.yaml
```diff
@@ -9,6 +9,11 @@ master:
   - AppFormer/uberfire: master
   - dashbuilder/dashbuilder: master
 
+7.4.x:
+  - errai/errai: 4.0.x
+  - AppFormer/uberfire: 1.4.x
+  - dashbuilder/dashbuilder: 1.0.x
```
psiroky: Does this mean that both 1.0.x and master now use the version 1.0.0-SNAPSHOT? We need to fix that I think.

mbiarnes: @psiroky this is fixed - dashbuilder master has to bumped up to 1.1.0-SNAPSHOT
look https://github.com/dashbuilder/dashbuilder/pull/379

psiroky: great, thanks

PR #571, opened 2017-10-10
added scripts for automated creation of jBPM-installer.zip and jBPM-i…

mbiarnes:
…nstaller-full.zip

(cherry picked from commit 5aea3c96ac29e6a9235e42cb19e29e5a0fd851b3)


psiroky:
Those scripts look very similar, duplicating a lot lines. Would it be possible to share the common parts e.g. with another "common" script? Or may we could have just single script doing both of the zips?

mbiarnes:
@psiroky will close this and redo a new one with should less conflictive

PR #577, opened 2017-10-16
override of version of xnio to align to EAP 7.0.8

mbiarnes:


ryanzhang:
@michaelneale I think that we should avoid adding this in droolsjbpm-build-bootstrap which will complex the version override. （There is a similar case to upgrade jbpmmigration to v0.16)
We should add these change into jboss-integration-platform-bom.
I understand that we might not do a release for jboss-integration-bom before LA. That is fine to me since this isn't a blocker issue.
Sorry if I don't make it clear enough before merging this PR. But I think we should revert this change to stabilize the CR build.
If you would do a new jboss-integration-platform, please don't forget to update xnio and jbpmmigrations.

manstis:
@mbiarnes Please see @ryanzhang comment above.

PR #583, opened 2017-10-26
DO NOT MERGE - RHBPMS-5026 - Version difference of jackson artifacts

mswiderski:
@mbiarnes @psiroky @krisv @sutaakar 

here are the changes to upgrade version of jackson to 2.8+ as it seems to be required on EAP 7.1

Though this won't work on lower versions like 7.0 or Wildfly 10 and 10.1. I tested this with Wildfly 11 and it does work as expected - tests where only covering what is reported in the jira

PR with code change in droolsjbpm-integration


psiroky:

mswiderski:
kie server build will fail on tests as it requires wildfly 11, so unless we upgrade to that version these will always be red. Do we have any plans for wildfly 11/eap 7.1?

psiroky:
@mswiderski if the jackson upgrade is required, then it seems we need to upgrade the WildFly as well. No plans for that at the moment though. Considering that it went final few days ago, we should start ASAP.

mareknovotny:
ok, I will merge this to not forget as this is anyway required to run on EAP 7.1/Wildfly 11, this will affect tests which uses widlfly10 profile and very probably fail on droolsjbpm-integration repository things.

PR #584, opened 2017-10-26
JDK9 build compatibility fixes

psiroky:
These changes should currently be enough for the simple `mvn clean install` build. We will likely need further fixes for tests - but that can be done separately.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -174,6 +174,8 @@
          IMPORTANT: always explain the reason for overriding the plugin version! -->
     <!-- Surefire version 2.19.1 coming from parent is buggy and kie-osgi tests in droolsjbpm-integration are failing because of that. -->
     <version.surefire.plugin>2.18.1</version.surefire.plugin>
+    <version.enforcer.plugin>3.0.0-M1</version.enforcer.plugin> <!-- Needed to support JDK 9. Can be removed once IP BOM 8.0.0.CR1 is used. -->
+    <version.javadoc.plugin>3.0.0-M1</version.javadoc.plugin> <!-- Needed to support JDK 9. Can be removed once IP BOM 8.0.0.CR1 is used. -->
     <!-- Jacoco plugin configurations -->
     <jacoco.haltOnFailure>false</jacoco.haltOnFailure>
     <jacoco.line.coveredratio.minimum>0.9</jacoco.line.coveredratio.minimum>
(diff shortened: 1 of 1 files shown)


psiroky:

psiroky:

psiroky:
I've proposed a fix for revapi: https://github.com/revapi/revapi/pull/99. We will need to wait for a new release unfortunately :(

psiroky:

psiroky:

pom.xml
```diff
@@ -1584,28 +1586,22 @@
       </activation>
       <properties>
         <!-- There is no public ECJ release (yet) which supports Java 9, so fallback to the native compiler -->
-        <!-- TODO remove the line below (and fix the surefire+failsafe config) after resolving https://issues.jboss.org/browse/DROOLS-1169 -->
-        <drools.dialect.java.compiler>NATIVE</drools.dialect.java.compiler>
```
ge0ffrey: Are you sure this line can be removed?
It's for running the tests, not compation and the related ECJ issue DROOLS-1169 isn't resolved yet.

pom.xml
```diff
@@ -174,6 +174,8 @@
          IMPORTANT: always explain the reason for overriding the plugin version! -->
     <!-- Surefire version 2.19.1 coming from parent is buggy and kie-osgi tests in droolsjbpm-integration are failing because of that. -->
     <version.surefire.plugin>2.18.1</version.surefire.plugin>
+    <version.enforcer.plugin>3.0.0-M1</version.enforcer.plugin> <!-- Needed to support JDK 9 -->
+    <version.javadoc.plugin>3.0.0-M1</version.javadoc.plugin> <!-- Needed to support JDK 9 -->
```
ge0ffrey: Can we add something in the comment as to when will we be able to remove these overrides?

We depend on ipbom 7.0 which extends jboss parent 22.
but ipbom 8.0 already extends jboss parent 25, which might already be JDK 9 compliant (as WildFly uses it too)

ge0ffrey:
Good to merge, only soft suggestions

pom.xml
psiroky: Yes, I just tried it locally and all works as expected with ECJ 4.4.2.

pom.xml
psiroky: I will add a comment, but we should find a better way to detect these. E.g. have a script that checks which property values are exactly the same as in parent, so can be safely removed. But that will of course require some time to implement.

psiroky:

PR #587, opened 2017-10-26
Latest released version is 7.4.1.Final

psiroky:


psiroky:

psiroky:
I've incorporated this change as part of https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/584 as it partially related to the revapi upgrades and should be easier to fix the build issues as part of the same PR.

PR #588, opened 2017-10-27
upgrade to java 9

mariofusco:


mariofusco:
DO NOT MERGE until mavel 2.4.0.Final won't be available on maven

pom.xml
```diff
@@ -2517,11 +2518,18 @@
...
         <groupId>org.mvel</groupId>
         <artifactId>mvel2</artifactId>
         <version>${version.org.mvel}</version>
       </dependency>
+
+      <dependency>
+        <groupId>javax.xml.bind</groupId>
+        <artifactId>jaxb-api</artifactId>
+        <version>${version.org.jaxb}</version>
+      </dependency>
```
psiroky: In ideal case, if possible, we should be using the JBoss spec jar (which already comes from parent no need to add it here):
```
      <dependency>
        <groupId>org.jboss.spec.javax.xml.bind</groupId>
        <artifactId>jboss-jaxb-api_2.2_spec</artifactId>
      </dependency>
```

ge0ffrey: Otherwise the jaxb spec jar will end up in the classpath twice for someone depending on, for example, this module and optaplanner-persistence-jaxb

mariofusco:
replaced by https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/593

PR #589, opened 2017-10-27
Gerrit master

mbiarnes:


mbiarnes:
This PR is for pushing the tags when doing a relese to Gerrit instead of jboss-integration

mbiarnes:
@psiroky can you please approve and merge this - when releasing I have to push the tags to Gerrit

psiroky:
Why are we not using the removed scripts? Or more precisely, how did we replace that functionality?

mbiarnes:
@psiroky these removed scripts are obsolete - I used them before we started with DSL. They are not used anymore.

psiroky:
I don't really understand why we don't need as them, as AFAICT we still need to e.g. create branches and push tags. But anyway, I no longer have the detailed insight into this, so will just trust your judgement.

script/release/kie-pushTag.sh
```diff
@@ -6,18 +6,23 @@ echo "target" : $target
 # clone the build-bootstrap that contains the other build scripts
 if [ "$target" == "community" ]; then
    source=kiegroup
+   remote=origin
 else
    source=jboss-integration
```
psiroky: Should we change also this? What does the `jboss-integration` mean ins this case?

mbiarnes:
Historically the tags were pushed to jboss-integration/<rep> - but now prod
guys asked if we can push the tags to Gerrit.
Aside from the tag there is a release branch we work with - so I push it to
kiegroup/<rep> if it is for community or jboss-integration/<rep> if it's
for productization.
Since prod wants to get rid of jboss-integration the question is, where to
push the branches. If in any case these branches should be pushed only to
kiegroup (i.e. r7.4.0.Final and bsync-7.4.x-2017.10.18).
But there is no decision made - I try my very best and I have to do parts
of the releases manually, because the reps change before the scripts.

PR #594, opened 2017-10-31
upgraded org.wildfly.core

mbiarnes:
makes only sense to merge this PR when https://github.com/jboss-integration/jboss-integration-platform-bom/pull/400 was merged and a new jboss-ip-bom was released


psiroky:
This should be clearly marked as DON'T MERGE, otherwise it may actually break the build. I would recommend closing the PR and raise one when upgrading the new IP BOM version. WDYT @mbiarnes?

mbiarnes:
you're right - I will close it

PR #595, opened 2017-11-01
Change version to check against with revapi to 7.0.0.Final

MarianMacik:
Hi, @psiroky,

here is the change we spoke about. So I put back the version we check against to 7.0.0.Final and removed version.org.kie.latestFinal.release property too since it is useless now. I have also gone through revapi-config.json files and updated their ignores and their comments so they are not confusing.

PR set:
- https://github.com/kiegroup/droolsjbpm-knowledge/pull/278
- https://github.com/kiegroup/jbpm/pull/1053
- https://github.com/kiegroup/drools/pull/1573
- https://github.com/kiegroup/droolsjbpm-integration/pull/1255
- https://github.com/kiegroup/optaplanner/pull/348

Thanks,

Marian

### DIFF
--- a/RELEASE-README.md
+++ b/RELEASE-README.md
@@ -426,10 +426,6 @@ If everything is perfect (compiles, Jenkins is all blue, sanity checks succeed a
            this is a property productisation needs to get the last released version on the branch where released from.
            When updated this should be pushed to the branch of the blessed repository
 
-        7. latest released *Final* version, property called `version.org.kie.latestFinal.release`
-           This property needs to be updated only with *Final* releases (not for Betas, CRs, etc). This property is being
-           used to check KIE API backwards compatibility (and there may be other uses as well).
-
     * Commit those changes (so you can tag them properly):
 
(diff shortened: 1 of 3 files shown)


psiroky:
@ryanzhang or @AntStephenson is the property `version.org.kie.latestFinal.release` used somewhere in the downstream prod. builds?

AntStephenson:
@psiroky - yes we override this for the last productized minor, afaik. If I recall correctly, its used in one of the projects for API checking between versions?

psiroky:
@AntStephenson it is no longer used for any API checks. However, we can keep the property around I believe. @MarianMacik could you please add the property back and set the version to 7.4.1.Final. We definitely need to improve how the property is being updated to make sure it is always up-to-date.

AntStephenson:
Afaik, we only use it for that, if its no longer used, we can just remove that property (actually I can do this tomorrow)?

psiroky:
@AntStephenson there are different properties for the API check: `revapi.oldKieVersion` and `revapi.newKieVersion`. Those should be used instead.

So @MarianMacik all is good, we don't need to add it back...

PR #610, opened 2017-11-14
changed URL to closed community releases $ removed script

mbiarnes:
As agreed with Kris it is a good idea to generate the jbpm-installers before releasing so people can download them and test them. Therefor are these changes.
- re-added the push of the jbpm-installers in script once we push all binaries to filemgmt.jboss.org
- changed the URL to the "closed" but not released Nexus repositories
- removed the script we had for pushing the jbpm-installers


mbiarnes:
Merge together with: https://github.com/kiegroup/kie-jenkins-scripts/pull/102

PR #625, opened 2017-11-28
JBPM-6637 - JDK9: fix failing jbpm tests

mswiderski:


pom.xml
```diff
@@ -249,6 +249,10 @@
     <version.org.jboss.spec.javax.websocket>1.0.0.Final</version.org.jboss.spec.javax.websocket>
 
     <version.org.mvel>2.4.0.Final</version.org.mvel>
+    
+    <version.javax.xml.soap>1.3.5</version.javax.xml.soap>
+    <version.javax.jws>1.0-MR1</version.javax.jws>
```
psiroky: Nit picking: the properties should be ideally added in alphabetic order...but considering that the POM is quite a mess already, this is not needed.

psiroky:

PR #626, opened 2017-11-28
added jbpm work items to repository list and to script to copy it to …

mswiderski:
…filemgmt server

@mbiarnes @psiroky could you please review and let me know what I did wrong?

@tsurdilo fyi as part of preparation for 7.5 community release

### DIFF
--- a/script/repository-list.txt
+++ b/script/repository-list.txt
@@ -14,3 +14,4 @@ jbpm-designer
 jbpm-wb
 kie-docs
 kie-wb-distributions
+jbpm-work-items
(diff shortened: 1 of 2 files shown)


script/repository-list.txt
```diff
@@ -14,3 +14,4 @@ jbpm-designer
 jbpm-wb
 kie-docs
 kie-wb-distributions
+jbpm-work-items
```
psiroky: I think this should go directly after the `jbpm` repo.

mswiderski: I put it as last one in as nothing else depends on it as the intention is whatever goes here should not be bundled with any app (other repository module). It should be purely available though service repository. So that way we could catch it before we release - as build should fail if something reference any work item from this repo. Does it make sense?

psiroky: Makes sense.

PR #627, opened 2017-11-28
AF-593+: Move uberfire to appformer.

mbarkley:
@psiroky Could you please review this? The primary goal of this PR set is to replace AppFormer/uberfire with kiegroup/appformer in the build (content being presently almost identical). I figgured it made sense to add kie-soup to the repositories list now too, but I don't have to do that here.


mbarkley:
Related:
https://github.com/AppFormer/uberfire/pull/927
https://github.com/kiegroup/appformer/pull/100
https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/627

PR #628, opened 2017-11-29
excluded appformer from update kie versions

mbiarnes:
since appformer will have it's own version

### DIFF
--- a/script/release/update-version-all.sh
+++ b/script/release/update-version-all.sh
@@ -89,7 +89,20 @@ for repository in `cat ${scriptDir}/../repository-list.txt` ; do
         echo "Repository: $repository"
         echo "==============================================================================="
         cd $repository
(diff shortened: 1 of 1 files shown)


psiroky:
@mbiarnes we will need to change the version, one way or the other.

mbiarnes:
@psiroky so the appformer(and dashbuilder) will have the same version as kie?

psiroky:
@mbiarnes ultimately yes. Not sure how long will that take though.

psiroky:
But regardless, we will need the ability to change the appformer versions (to whatever value will be chosen).

mbiarnes:
@psiroky ths is a workaround! we have to think about something more stable. 
we could change the update scripts to use also a appformer version. since kie-soup and kie will always have the same version the only second variable will be the version for appformer.
I only know that the nightly builds will crash until this PR is merged (as workaround) or a more stable PR is raised

PR #636, opened 2017-12-02
tweaked update-all-versions for drlx-parser

mbiarnes:
MERGE makes only sense if drlx-parser is part of repository_list.txt

### DIFF
--- a/script/release/update-version-all.sh
+++ b/script/release/update-version-all.sh
@@ -128,6 +128,13 @@ for repository in `cat ${scriptDir}/../repository-list.txt` ; do
             mvn -B -s $settingsXmlFile clean install -DskipTests
             returnCode=$?
 
+        elif [ "$repository" = "drlx-parser" ];then
+            #update version in drlx-parser/pom.xml
+            cd drlx-parser
+            mvnVersionsUpdateParent
+            cd ..
+            returnCode=$?
+
         elif [ "$repository" = "jbpm" ]; then
             mvnVersionsUpdateParentAndChildModules
             returnCode=$?


mbiarnes:
@psiroky like this?

mbiarnes:
@psiroky so - that doesn't work on my local:
mvn -B -N -e versions:update-parent -Dfull -DparentVersion="[$newVersion]" -DallowSnapshots=true -DgenerateBackupPoms=false 
with newVersion=7.6.0.Final
[WARNING] Not updating version: could not resolve any versions

mbiarnes:
@psiroky I see - drlx-parser has still 7.5.0-SNAPSHOT. So i will change this and retest this

mbiarnes:
@psiroky 
https://gist.github.com/mbiarnes/058191f5ed8dd7d01292d0fde2f0e86a

psiroky:
`[WARNING] Not updating version: could not resolve any versions` -- what version are you upgrading to?

mbiarnes:
@psiroky for testing: newVersion=7.6.0.Final

psiroky:
@mbiarnes I think the reason for the error is that the parent POM actually does not exist in that version. You would need to build it locally. Or perhaps try with some version that is already available in local/remote repo (like 7.4.1.Final).

mbiarnes:
@psiroky doesn't work:
- the pom has the version 7.6.0-SNAPSHOT
- build project so the pom is in my .m2
- tried to update parent --> error
- changed version of pom to 7.4.1.Final
- build project
- tried to update parent --> error
what am i missng?

psiroky:
@mbiarnes the `newVersion` needs to be available. Leave the POM with `7.6.0-SNAPSHOT` and set `newVersion=7.4.1.Final`.

mbiarnes:
@psiroky OK - with this it works. 
What  am missing is: we want to upgrade to a new version (update-all-version.sh) - 
Must the drlx-parser version exist before we upgrade kie to the "same" version?

psiroky:
@mbiarnes I don't think so. However, it is quite easy to test that. Just update the repository-list.txt locally and run the update-version-all script on all repos.

mbiarnes:
@psiroky OK works. But this PR #543 should be merged too.

PR #637, opened 2017-12-02
fixed update-all-version.sh incl. kie-soup and appformer

mbiarnes:
version change also for kie-soup and appformer without skipping these reps


mbiarnes:
MERGE together with https://github.com/kiegroup/kie-jenkins-scripts/pull/118

PR #642, opened 2017-12-11
upgraded appformer version

mbiarnes:
during kie 7.5.0.Final release appformer was upgraded by error to 2.2.0.Final instead of 2.1.0.Final. So 2.2.0.Final is existing. We should upgrade the appformer develop version to 2.2.1-SNAPSHOT


mbiarnes:
Merge after https://github.com/kiegroup/appformer/pull/114 was merged

PR #673, opened 2018-01-17
Add Guice to pom.xml and remove all of its exclusions.

kkufova:
This PR should fix the failing tests in https://github.com/kiegroup/appformer/pull/126 caused by `java.lang.NoClassDefFoundError: com/google/inject/Module`.

The https://github.com/kiegroup/appformer/pull/126 PR changes the parent of AppFormer from `jboss-integration-platform-bom` to `kie-parent`.

Another linked PR: https://github.com/jboss-integration/jboss-integration-platform-bom/pull/414.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -128,6 +128,7 @@
     <version.com.google.gwt>2.8.2</version.com.google.gwt>
     <version.net.ltgt.gwt.maven>1.0-rc-6</version.net.ltgt.gwt.maven>
     <version.com.google.guava>20.0</version.com.google.guava>
+    <version.com.google.inject.guice>4.0</version.com.google.inject.guice>
     <!-- this version will overwrite jboss-ip-version 3.1.2. The version in jboss-ip-bom couldn't be upgraded as this version needs GWT 2.7.0 -->
     <!-- please look at https://github.com/fredsa/gwt-dnd/wiki/GettingStarted -->
     <version.com.allen-sauer.gwt.dnd>3.3.3</version.com.allen-sauer.gwt.dnd>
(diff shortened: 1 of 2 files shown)


kkufova:

kkufova:

manstis:
@kkufova IDK if it's at all related; but we've had problems with Plexus in a CDI environment before. See 
https://github.com/kiegroup/kie-soup/blob/master/kie-soup-maven-utils/kie-soup-maven-integration/src/main/java/org/appformer/maven/integration/DependentScopeModule.java

I also note we exclude guice from many places; so not sure adding it back is the right answer?

See https://github.com/kiegroup/appformer/blob/master/uberfire-m2repo-editor/uberfire-m2repo-editor-backend/pom.xml (and search for guice).

Do you know why ```com.google.inject.Module``` is needed?

manstis:
@kkufova BTW I'm not saying there may not be a legitimate reason to add it back; but think the proposal should be considered by a wider audience as it's impact could be further reaching than simply making the failing tests pass.

@mariofusco @porcelli @ederign Anybody recall what we do with Guice and any problems?

jomarko:
@manstis do you expect hidden problems even if the full downstream build pass? I mean like problems that appear in runtime?

manstis:
@jomarko IIRC yes. However if the "extended QE" test suite passes too; then we can probably _feel_ more confident. However I raised the alarm as I have a feeling it could cause problems. Others may feel otherwise and we can all move on.. and my credibility can take a step backwards ;-)

kkufova:
@manstis, Guice was excluded from `droolsjbpm-build-bootstrap` by Petr Siroky. I asked him why, and he replied that he does not really remember the reason and that it is possible that the reason is no longer valid and that I could try to add it back. `com.google.inject.Module` is needed because of these failures: 
`java.lang.NoClassDefFoundError: com/google/inject/Module` in https://github.com/kiegroup/appformer/pull/126.

porcelli:
we need @mbarkley to look into this

psiroky:
FWIW I am _not_ the original author of the exclusions. My name appears in the `git blame` only because I was shuffling things around. This may be the original commit (or one of them): https://github.com/kiegroup/droolsjbpm-build-bootstrap/commit/54e21bf745f3f0b9936bcefec0beacd6519f7839

PR #683, opened 2018-01-23
Add --target-repo option to git-all.sh and mvn-all.sh

yurloc:
With this option I can run `mvn-all.sh clean install --target-repo=drools-wb` and only build the repositories that `drools-wb` depends on:
* kie-soup
* appformer
* droolsjbpm-build-bootstrap
* drlx-parser
* droolsjbpm-knowledge
* drools
* jbpm
* optaplanner
* droolsjbpm-integration
* kie-wb-playground
* kie-wb-common
* drools-wb

which is 12 out of 20. This can be handy when building a local feature that spans multiple repositories. See the full dependency graph generated by the `repo-dep-tree.pl` script.

dep-tree dot

### DIFF
--- a/README.md
+++ b/README.md
@@ -416,6 +416,11 @@ Running the build
 
     * Note: the `mvn-all.sh` script is working directory independent.
 
+* You can use `mvn-all.sh` to compile a specific repository and all repositories that your target repository depends on.
+  This is done using the `--target-repo` option which will invoke `repo-dep-tree.pl` script to discover cross-repository
+  project dependencies. Use `--repo-list` to specify custom list of repositories. These options work for `git-all.sh`
+  too.
+
 * Warning: The first `mvn` build of a day will download the latest SNAPSHOT dependencies of other kiegroup projects,
 unless you build all those kiegroup projects from source.
 Those SNAPSHOTS were build and deployed last night by Jenkins jobs.
--- a/script/checks/repo-dep-tree.pl
+++ b/script/checks/repo-dep-tree.pl
@@ -49,13 +49,13 @@ if( exists $opts{d} ) {
 }
 
 if( exists $opts{h} ) {
-  print "\n";
-  print " h ........ print [h]elp (this)\n";
-  print " d ........ create [d]ot file (implies -f)\n";
-  print " f ........ [f]ilter transitive dependencies (not compatible with -t)\n";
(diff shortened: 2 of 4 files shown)


mareknovotny:
@yurloc I can't run it with the new option (without that option it runs on all repos)

For instance:
```
droolsjbpm-build-bootstrap/script/git-all.sh fetch --target-repo=drools
Git arg. line=fetch

Total time: 0s
```
Any idea what could be wrong or am I doing something wrong?

yurloc:
@mareknovotny The `repo-dep-tree.pl` failed for some reason. Could be a missing repository on your FS or something else. We don't know because I redirected error output to `/dev/null` which was a bad idea. It's fixed now so if you try again, you should find out why `repo-dep-tree.pl` failed. If the reason is that one of the repositories listed in `droolsjbpm-build-bootstrap/scripts/repository-list.txt` is missing, then the question is if it should be skipped like `git-all` and `mvn-all` scripts do.

mareknovotny:
thanks @yurloc, I now get 
```
Can't locate XML/Simple.pm in @INC (you may need to install the XML::Simple module) (@INC contains: /usr/local/lib64/perl5 /usr/local/share/perl5 /usr/lib64/perl5/vendor_perl /usr/share/perl5/vendor_perl /usr/lib64/perl5 /usr/share/perl5) at /home/mnovotny/projects/git/KIEGROUP/droolsjbpm-build-bootstrap/script/checks/repo-dep-tree.pl line 23.
BEGIN failed--compilation aborted at /home/mnovotny/projects/git/KIEGROUP/droolsjbpm-build-bootstrap/script/checks/repo-dep-tree.pl line 23.
Git arg. line=fetch

Total time: 0s
```
I need to find out why it can't be located as it seems I have installed it
```
dnf install perl-XML-Parser
...
Package perl-XML-Parser-2.44-9.fc27.x86_64 is already installed, skipping.
```

mareknovotny:
ok, I've made it to work with configuring and installing through cpan shell. I don't know why dnf doesn't setup the perl library path correctly.

One thing, do you want to update README.md too with this ability?

yurloc:
@mareknovotny Yes, that's a good idea. I'll try to cover the `--repo-list` option too since it's missing as well.

yurloc:
@mareknovotny Readme updated.

PR #693, opened 2018-02-06
[DO NOT MERGE] JBPM-6806: Included KIE Lienzo artifacts into the dependency management.

romartin:
Hey @mbiarnes @tiagodolphine @manstis 

Created ticket and branches with same names for doing the right Jenkins integration tests. 
See https://issues.jboss.org/browse/JBPM-6806

Please **DO NOT MERGE** yet.

Thanks!

### DIFF
--- a/kie-bom/pom.xml
+++ b/kie-bom/pom.xml
@@ -538,6 +538,31 @@
         <classifier>sources</classifier>
       </dependency>
 
+      <!-- KIE - Custom lienzo artifacts -->
+      <dependency>
+        <groupId>org.kie</groupId>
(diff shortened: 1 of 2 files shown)


mbiarnes:
Jenkins retest this please

mbiarnes:
@romartin wouldn't it be easier if we add lienzo-core and lienzo-test to the repository-list.txt?
So it builds always before all other projects.
It will release automatically with all other reps.
Only the update-versions-all script has to be changed.

romartin:
@mbiarnes yes I see you created a PR for this, but anyway I think this is still necessary.

PR #733, opened 2018-04-09
Upgrade maven version to 3.5.0 for productized maven

ryanzhang:
We need to productized the multiple maven components.
So request to upgrade 3.5.0 to match the productized maven components.


ryanzhang:

mbiarnes:
@ryanzhang @mareknovotny didn't we agree to go for 3.5.2 ?
wasn't 3.5.0 buggy?

ryanzhang:
@mbiarnes  There is some compile error because maven-aether-provider is removed in 3.5.0. Probably there are more change happen during 3.3.9 and 3.5.0
Let's hold on this.

PR #740, opened 2018-04-20
RHPAM-340: resteasy upgrade

mbiarnes:
(cherry picked from commit 38beaf2e1f5d1fa209c17e960fe4b900a5689159)


mareknovotny:

mbiarnes:

mbiarnes:
closed because:
There is a bug in the kie-jenkins-plugin / GitHub API where it occasionally returns null data, this then gets cached in the build and simply restarting the Jenkins build won't fix this. This phenomenon seems to occur very infrequently and we've not had much in the way of spare time, so we've not scheduled any work to add a retry mechanism to the plugin (there is already an issue raised for it).

The fix right now is to abandon this PR and raise it again. The kie-jenkins-plugin will fetch new data

PR #741, opened 2018-04-26
RHPAM-340: resteasy upgrade

mbiarnes:


mbiarnes:

mbiarnes:
@krisv @mareknovotny @ryanzhang 
I don't see the  a direct correlation to a new version of resteasy and the failing tests.

krisv:

krisv:
@mswiderski could these kie-server test failures be related to this resteasy upgrade?

mswiderski:
kie server does not depend on rest easy directly but through the app server it is deployed to. So these test failures are not related but to be able to completely verify that kie server tests should b executed on tomcat as it the uses RESTEasy as rest provider directly.

Is this upgrade aligned with WildFly/EAP version? If so what is the version it is aligned with?

mareknovotny:
the latest wildfly uses 3.5.0.Final but what is in EAP is 3.0.25.Final

PR #757, opened 2018-05-16
RHPAM-1017 Add freemarker into dependencyMgmt

ryanzhang:
Freemarker needs to be added into dependencyMgmt so that product build can override the dependency version to productize version.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2678,6 +2678,12 @@
         <version>${version.javax.xml.soap}</version>
       </dependency>
 
+      <dependency>
+        <groupId>org.freemarker</groupId>
+        <artifactId>freemarker</artifactId>
+        <version>${version.org.freemarker}</version>
+      </dependency>
+
     </dependencies>
 
   </dependencyManagement>


mareknovotny:
Why do we need to manage it in droolsjbpm-build-bootstrap when it is in https://github.com/jboss-integration/jboss-integration-platform-bom/blob/master/ip-bom/pom.xml#L1795-L1799 ? The property definition should be enough in  https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/pom.xml#L96 or not?

ryanzhang:
When doing productization, the PME(maven-extension) would override the community version to product version when it's an explicit managed dependency. 
In this case,  we don't want to release new jb-ip-bom's for override freemarker. This could cause more risk.
Instead, we want to override freemarker's version in kie-parent, so that the freemarker needs to explicit to be declared here.

mareknovotny:
@ryanzhang I get that PME overriden, so I was wondering just that we are updating the community kie-parent when the IP BOM is defining the dependency management and later we would need additional change to fix all duplications of dependency management. So this change is more about hacking the definition. Isn't better to temporary patch it only on gerrit's source?

PR #766, opened 2018-06-13
AF-1243: Add new drools-wb-ui artifact

yurloc:
Related PRs:
kiegroup/droolsjbpm-build-bootstrap#766
kiegroup/drools-wb#873

The new artifact is a result of splitting drools-wb-webapp into two
modules, which is a standard layout with the new tbroyer GWT plugin.

### DIFF
--- a/drools-bom/pom.xml
+++ b/drools-bom/pom.xml
@@ -1088,6 +1088,12 @@
       <!--drools-wb distribution-->
       <dependency>
         <groupId>org.drools</groupId>
+        <artifactId>drools-wb-ui</artifactId>
+        <type>war</type>
+        <version>${version.org.kie}</version>
+      </dependency>
+      <dependency>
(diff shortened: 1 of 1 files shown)


drools-bom/pom.xml
```diff
@@ -1086,6 +1086,12 @@
       <!--REST-->
 
       <!--drools-wb distribution-->
+      <dependency>
+        <groupId>org.drools</groupId>
+        <artifactId>drools-wb-ui</artifactId>
```
rsynek: Could you please explain what's the purpose of this new artifact?

rsynek: answered: the reason is build process - ui module contains client code, webapp backend + descriptors.

ge0ffrey:
Merged this already, without merging the related PR as this works as is already.

PR #767, opened 2018-06-15
JBPM-7372 - single zip distribution for jBPM

mswiderski:
@krisv @tsurdilo @mbiarnes here is the single zip jbpm server distribution


mbiarnes:
@tsurdilo @mswiderski @krisv I did the test with both PR (one droolsjbpm-build-boostrap and kie-wb-distributions) and it worked properly. No test failures.

PR #771, opened 2018-07-04
BAQE-439 Introduce dependency on Tomcat DBCP for DB connection pooling

rsynek:
@AntStephenson could you please review?


AntStephenson:
Approved pending a green build

rsynek:
@AntStephenson increased Narayana version to the latest release - makes sense to do it as a part of 
BAQE-439.

ryanzhang:
For some reason, nightly build start to pull org.apache.tomcat:catalina:jar:9.0.10 which does not exist?
What should be the expected tomcat catalina version? should we define it as well?

ryanzhang:
I put another PR to fix the issue I mentioned https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/773

PR #773, opened 2018-07-11
Fix tomcat-dbcp properties to avoid overriding catalina version

ryanzhang:
org.apache.tomcat.catalina also used maven property version.org.apache.tomcat, so rename the properties to avoid the unwanted override.


rsynek:
@ryanzhang thanks Ryan for the fix - and apologies, I haven't checked the IP bom first.

ryanzhang:
Not a problem, @rsynek 
I am going to merge it and rebuild the master branch.

PR #775, opened 2018-07-11
Change revapi to check against 7.5.0.Final

MarianMacik:
Wait with merge until all valid repositories are updated.

@rsynek 

We now have to check against latest product release (7.0 DM/PAM) which means that we have to check against 7.5.0.Final community release (not 7.7.0.Final since that won't contain DM release).


ge0ffrey:
All merged.

PR #801, opened 2018-08-23
JBPM-7635 - Upgrade to WildFly 13

mswiderski:


mareknovotny:
we should use Wildfly 14.0.1 as the latest possible release and there is already a PR for that even it is far away from what we should update there. So these changes and dependending PRs could be good to combine and synchronize on that version. Also the important other deps to not cause dependency clashes.

mswiderski:
this one is out of date, closing

PR #805, opened 2018-08-28
Upgrade tomcat-dbcp to 9.0.11 to consume fixes

rsynek:
@mareknovotny could you please take a look?


mareknovotny:
this is independent on any of your PRs, right? so we can merge it when we are ok to do.

PR #814, opened 2018-09-04
upgrade to mvel 2.4.2.Final

mariofusco:


mariofusco:
Version of mvel 2.4.1.Final caused a regression in type coercion. The regression will be fixed in 2.4.2.Final. I'm leaving this PR open and I'll upgrade it to that version when it will be available on maven central.

mariofusco:

PR #817, opened 2018-09-06
[DROOLS-2976] Add build-helper-maven-plugin as managed version

gitgabrio:
@manstis @jomarko 
Scope of this PR is to add builder-helper-maven-plugin as managed plugin, to be used inside drools-wb-webapp

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1338,6 +1338,11 @@
           <artifactId>wildfly-maven-plugin</artifactId>
           <version>1.2.0.Final</version>
         </plugin>
+        <plugin>
+          <groupId>org.codehaus.mojo</groupId>
+          <artifactId>build-helper-maven-plugin</artifactId>
+          <version>3.0.0</version>
+        </plugin>
       </plugins>
     </pluginManagement>
 


pom.xml
```diff
@@ -1338,6 +1338,11 @@
           <artifactId>wildfly-maven-plugin</artifactId>
           <version>1.2.0.Final</version>
         </plugin>
+        <plugin>
```
manstis: Actually, I compiled ```drools-wb``` locally (with your PR and without this PR) and it succeeded. 

So, on investigation, I found [this](https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/pom.xml#L956). Do we need to re-declare the plugin or simply add the version to the existing entry?

gitgabrio: @manstis I will add the version to the existing entry and remove the new declaration.
Fixed with
[DROOLS-2976] Add version to existing build-helper-maven-plugin declaration. Removing duplicated one

manstis:
@gitgabrio Oeeee.. a merge conflict!?!

gitgabrio:
@manstis could you merge this one please?

PR #819, opened 2018-09-07
change kie-docs location and fix for images build

mramendi:
Hi Michael,

The location of community docs has changed in the kie-docs repo. Also, we have just found an issue with images not copying and wrong links to images in docs. While we look for a more proper fix inside Maven. I wrote a post-processing script fixing the issue after the build, so that we would not hold back release.

Please do test this in your Jenkins environment; I did test the script manually.


mbiarnes:
@mramendi @sterobin @jomarko will close this PR since we don't need it any more

PR #825, opened 2018-09-10
updated Maven min version

mbiarnes:


mareknovotny:
why don't we upgrade to the latest available 3.5.4?

mareknovotny:
just curious if minimal version is somewhere described as a limitation for a feature or it is just a good to have as minimum

PR #829, opened 2018-09-13
[AF-1528] 7.11.x Workbench offline

desmax74:
Workbench offline 7.11.x
https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/829
https://github.com/kiegroup/kie-soup/pull/57
https://github.com/kiegroup/appformer/pull/459
https://github.com/kiegroup/kie-wb-common/pull/2117
https://github.com/kiegroup/kie-wb-distributions/pull/811

@porcelli @domhanak @jhrcek

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2690,19 +2690,16 @@
         <groupId>org.postgresql</groupId>
         <artifactId>postgresql</artifactId>
         <version>${version.org.postgresql}</version>
-        <scope>runtime</scope>
       </dependency>
       <dependency>
         <groupId>mysql</groupId>
         <artifactId>mysql-connector-java</artifactId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -2682,6 +2685,26 @@
         <artifactId>ecj</artifactId>
         <version>${version.ecj}</version>
       </dependency>
+
+      <dependency>
```
jhrcek: This is `<dependencyManagement>` so we should not be declaring scopes here (just defining versions). Scope should be defined at the level of individual poms that actually use these in their `<dependenciese>` section. Please remove the `<scope>runtime</scope>` from all three deps..

desmax74: Fixed was copy and paste

PR #833, opened 2018-09-17
Updated to 8.2.0-SNAPSHOT to use protobuf 3.6.1

lucamolteni:
.


lucamolteni:
@mbiarnes Ok so can you release Final with updated protobuf? In the meanwhile I'll push only the bump

mbiarnes:
@lucamolteni we should close this PR - the update of protobuf in jboss-ip-bom is merged; asap as we do a new release of jboss-ip-bom 8.2.0.Final and we upgrade our kie-parent we have this version. In case you cant wait you should do an overwrite ih kie-parent

lucamolteni:
@mbiarnes Perfect fine by me

Thank you

PR #844, opened 2018-09-26
JDK11: Add dependencies needed by droolsjbpm-integration/kie-camel

cimbalek:
...to compile and run with JDK11

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -247,6 +247,8 @@
     <version.org.jboss.spec.javax.websocket>1.1.1.Final</version.org.jboss.spec.javax.websocket>
 
     <version.javax.xml.soap>1.3.5</version.javax.xml.soap>
+    <version.com.sun.xml.messaging.saaj-impl>1.4.0</version.com.sun.xml.messaging.saaj-impl>
+    <version.javax.xml.ws>2.3.0</version.javax.xml.ws>
     <version.javax.jws>1.0-MR1</version.javax.jws>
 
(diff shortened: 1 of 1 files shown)


pszubiak:

mareknovotny:
This should be tested with openjdk 11 then on CI
Do we need to setup special slave with that version?

pszubiak:
@mareknovotny Not yet. And we use Oracle JDK instead of OpenJDK. Adding Oracle JDK 11 should be quite easy. Probably we should switch to OpenJDK too, but it will require more work/tests.

cimbalek:
I guess we need to merge this first and then possibly resolve Java 11 issues :) Important is that this breaks nothing for Java 8

mareknovotny:
There is some enforcer ban-duplicated-classes rule errors

cimbalek:
Who should I ask about these enforcer rules what to do with them? We need to resolve it somehow. Sooner = better

I guess we cannot avoid duplicities if something is present in JDK8 and is not in JDK11, correct me if I'm wrong

mareknovotny:
We need to sort them out. Can you work on exclusions or redefinition of the dependencies to not duplicate classes? 
I haven't analysed the real duplication, but this addition reasonably adding some duplication if the enforcer reports them.

mareknovotny:
can we close this PR or is it for rebase and update?

cimbalek:
close it. I think it is solved already by different PR

PR #847, opened 2018-10-05
Jboss ip bom 8.2.0.final

mbiarnes:

PR #848, opened 2018-10-09
Updated Parent to 8.2.0 Final to use new version of protobuf

lucamolteni:
@mbiarnes Is this correct?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2456,7 +2456,7 @@
         <artifactId>zjsonpatch</artifactId>
         <version>${version.zjsonpatch}</version>
       </dependency>
- 
+
       <dependency>
         <groupId>org.slf4j</groupId>
         <artifactId>jul-to-slf4j</artifactId>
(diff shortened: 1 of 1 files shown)


mbiarnes:
@lu

> @mbiarnes Is this correct?

yes and no :) please look at PR #847 - there is the same update + a migration of a dependency to the jboss-ip-bom 8.2.0.Final. Furthermore the jboss-ip-bom version should be changes in pom.xml and kie-user-parent-bom/pom.xml :) Since this PR does what #847 does - this PR can be closed.

lucamolteni:
@mbiarnes Ok but still I need this branch name to execute the tests on drools project so I don't think we should close this already?

mbiarnes:
@lucamolteni  fyi - all PRs related the upgrade of jboss-ip-bom 8.2.0.Final wre merged yesterday. 
The version used there is https://github.com/jboss-integration/jboss-integration-platform-bom/blob/8.2.0.Final/pom.xml#L139
As soon as you can please close this PR.

PR #877, opened 2018-11-05
added missing dependency cxf-rt-rs-client

mbiarnes:
added missing dependency


mareknovotny:
just a head up, I ran the tests in kie-camel and it still shows https://pastebin.com/cdb3m0m5

PR #882, opened 2018-11-08
removed mvel2 as it is in jboss-ip-bom

mbiarnes:


mbiarnes:
this dependency is already in the [jboss-ip-bom](https://github.com/jboss-integration/jboss-integration-platform-bom/blob/8.3.2.Final/ip-bom/pom.xml#L2755)

PR #883, opened 2018-11-09
[DROOLS-3194] jdk11 support

mariofusco:


mariofusco:

PR #897, opened 2018-12-03
[JBPM-8019] add dependencies necessary for jbpm to run on jdk 11

mariofusco:


pom.xml
```diff
@@ -2702,6 +2702,36 @@
...
+      </dependency>
+      <dependency>
+        <groupId>javax.annotation</groupId>
+        <artifactId>javax.annotation-api</artifactId>
+        <version>1.3.2</version>
+      </dependency>
+      <dependency>
+        <groupId>javax.xml.soap</groupId>
+        <artifactId>javax.xml.soap-api</artifactId>
+        <version>1.4.0</version>
```
yurloc: This is a duplicate dependency! The dependency on `javax.xml.soap-api` is already declared in `dependencyManagement` with `version.javax.xml.soap` which is 1.3.5. Please remove one of the dependencies.

PR #899, opened 2018-12-10
[RHDM-709][RHPAM-1453] upgrade mvel version

mariofusco:


mariofusco:

pom.xml
```diff
@@ -261,6 +261,7 @@
     <version.org.wildfly>14.0.1.Final</version.org.wildfly>
     <version.org.wildfly.core>6.0.2.Final</version.org.wildfly.core>
 
+    <version.org.mvel>2.4.3.Final</version.org.mvel>
```
mareknovotny: we have mvel version property in IP BOM https://github.com/jboss-integration/jboss-integration-platform-bom/blob/8.3.2.Final/pom.xml#L326.
@mariofusco @baldimir use that dependency management. Adding such properties will confuse people.

tagging here @mbiarnes

mbiarnes:
the upgrade of version.org.mavel was done in this PR https://github.com/jboss-integration/jboss-integration-platform-bom/pull/442 - once released a new jboss-ip-bom 8.3.3.Final for kie 7.14.x and upgraded to use it we can close this PR. The version <version.org.mvel>2.4.3.Final will be then inherited from jboss-ip-bom.

PR #910, opened 2019-01-11
added missing setting.xml to upgrade-version script modified default …

mbiarnes:
…parameter when creating release branches

### DIFF
--- a/script/release/kie-createReleaseBranches.sh
+++ b/script/release/kie-createReleaseBranches.sh
@@ -37,7 +37,7 @@ if [ "$source" == "production-tag" ]; then
 fi
 
 # upgrades the version to the release/tag version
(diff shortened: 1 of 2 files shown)


script/release/kie-createReleaseBranches.sh
```diff
@@ -37,7 +37,7 @@ if [ "$source" == "production-tag" ]; then
 fi
 
 # upgrades the version to the release/tag version
-./droolsjbpm-build-bootstrap/script/release/update-version-all.sh $releaseVersion $uberfireVersion $target
+./droolsjbpm-build-bootstrap/script/release/update-version-all.sh $releaseVersion $uberfireVersion custom
```
mareknovotny: what is that `custom` argument?

PR #922, opened 2019-02-05
added dependency javax.xml.soap-api to avoid jbpm failing

mbiarnes:


mbiarnes:
@yurloc requested changes done

PR #931, opened 2019-02-20
Drools Constraint Parser

lucamolteni:


mareknovotny:
just I went through this and noticed you haven't removed drlx-parser repository from https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/script/repository-list.txt#L7, is that overlooked?

mareknovotny:
fyi @mbiarnes  @cimbalek, @pszubiak, @AntStephenson  we need to project the dropping of drlx-parser on master into KIE CI job definitions too

lucamolteni:
> just I went through this and noticed you haven't removed drlx-parser repository from https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/script/repository-list.txt#L7, is that overlooked?

Yes sorry I will do it in the following days

PR #965, opened 2019-05-31
Add JMH version

Fonifan:


pom.xml
```diff
@@ -155,6 +155,9 @@
     <!-- override of version coming from jboss-ip-bom since this is needed for kie-wb-common -->
     <version.org.mockito>1.10.19</version.org.mockito>
 
+    <!-- Version of JMH -->
+    <version.org.openjdk.jmh>1.21</version.org.openjdk.jmh>
```
rsynek: One suggestion: to define here also dependency management for JMH, so that it does not have to repeat over different repositories.

Fonifan: Done.

rsynek:
jenkins execute compile downstream build

PR #974, opened 2019-06-26
needed override of com.thoughtworks.xstream version

mbiarnes:


mbiarnes:

pom.xml
```diff
@@ -265,6 +265,9 @@
     <version.com.github.javaparser>3.13.10</version.com.github.javaparser>
 
     <version.org.apache.commons.csv>1.6</version.org.apache.commons.csv>
+
+    <!-- override needed for 7.4.0.CR2 -->
```
mareknovotny: please don't add such comments into upstream here ;)

mareknovotny:

pom.xml
mbiarnes: when the full downstream build is green. before merging this I will remove this comment

mareknovotny:
the failed tests now from 2nd run are different, so I would say it is ok.
Please send PR for master too.

krisv:
@mbiarnes can you make sure the same gets fixed on master too (although with ip-bom getting removed it's going to be slightly different location I guess).

mbiarnes:
@krisv already done: https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/pom.xml#L108

PR #979, opened 2019-07-08
BAQE-1001 Unify reporting to SonarCloud under KIE parent

rsynek:
- jacoco:report, jacoco:merge and sonar:sonar are configured in pom.xml instead of triggering them directly
- sonarcloud-related properties are read from environment variables that are set by Jenkins
- PR analysis is moved to a separate profile

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1702,6 +1702,11 @@
           <artifactId>wildfly-maven-plugin</artifactId>
           <version>1.2.0.Final</version>
         </plugin>
+        <plugin>
+          <groupId>org.sonarsource.scanner.maven</groupId>
+          <artifactId>sonar-maven-plugin</artifactId>
+          <version>3.6.0.1398</version>
+        </plugin>
       </plugins>
     </pluginManagement>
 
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -2044,6 +2049,86 @@
...
+      <properties>
+        <sonar.pullrequest.provider>GitHub</sonar.pullrequest.provider>
+        <!--suppress UnresolvedMavenProperty -->
+        <sonar.pullrequest.branch>${env.ghprbSourceBranch}</sonar.pullrequest.branch>
+        <!--suppress UnresolvedMavenProperty -->
+        <sonar.pullrequest.key>${env.ghprbPullId}</sonar.pullrequest.key>
+        <!--suppress UnresolvedMavenProperty -->
+        <sonar.pullrequest.base>${env.ghprbTargetBranch}</sonar.pullrequest.base>
+        <!--suppress UnresolvedMavenProperty -->
+        <sonar.pullrequest.github.repository>${env.ghprbAuthorRepoGitUrl}</sonar.pullrequest.github.repository>
```
yurloc: https://sonarcloud.io/documentation/integrations/github/ says:
> sonar.pullrequest.github.repository=my-company/my-repo

So after looking at Jenkins properties, I'd say it should be `${env.ghprbGhRepository}`:
Screen Shot 2019-07-11 at 13 00 29

Wdyt?

rsynek: Interestingly, here https://sonarcloud.io/documentation/analysis/pull-request/ they don't mention the property at all.

The analysis works - before moving this profile to kie-parent, the same environment variable was passed to sonar scanner via a property:
https://github.com/kiegroup/kie-jenkins-scripts/pull/492/files#diff-62de59f2b1d8b83097e60b2792cb1599L261

If I take a look at vehicle routing, there is no such property and the analysis works too. 

I think it's not required for newly added projects, that are paired with corresponding github repositories.

yurloc: Yes, it's probably redundant for our projects. Shall we remove it?

PR #984, opened 2019-07-11
fixed update script and kie-createReleaseBranches.sh

mbiarnes:
version upgrades that doesn't update automatically and have to be changed "manually" now are changed directly in the update-version-all.sh script - so this is not needed any more in kie-createReleaseBranches.sh  script

With these changes the update of versions work:
- it respects the new kie-soup-bom and uberfire-bom in droolsjbpm-build-bootstrap
- is respects the own version number of uberfire


mbiarnes:
@krisv @danielezonca can you please check this - it is needed for releases

mbiarnes:
> We should do an actual test release as well, but changes look ok.

will do so :)

PR #987, opened 2019-07-12
Return failsafe plugin back to default plugin executions

sutaakar:
Failsafe plugin was unintentionally removed in https://github.com/kiegroup/droolsjbpm-build-bootstrap/commit/fc094e3b4d14edde0a09c4de95f01ee5e8902326#diff-600376dffeb79835ede4a0b285078036L1415 .
The result is that majority of integration tests were not executed since then.


sutaakar:

mbiarnes:

PR #998, opened 2019-07-19
added exclusion for banned dependency

mbiarnes:
kieAllBuild-master failed because of 
22:53:16 [WARNING] Rule 0: org.apache.maven.plugins.enforcer.BannedDependencies failed with message:
22:53:16 Found Banned Dependency: commons-logging:commons-logging:jar:1.1.1
22:53:16 Use 'mvn dependency:tree' to locate the source of the banned dependencies

commons-logging is brought in as transitive dependency from resteasy-jaxrs.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4467,6 +4467,12 @@
         <groupId>org.jboss.resteasy</groupId>
         <artifactId>resteasy-jaxrs</artifactId>
         <version>${version.org.jboss.resteasy}</version>
+        <exclusions>
+          <exclusion>
+            <groupId>commons-logging</groupId>
+            <artifactId>commons-logging</artifactId>
+          </exclusion>
+        </exclusions>
       </dependency>
       <dependency>
         <groupId>org.jboss.resteasy</groupId>


wmedvede:

wmedvede:
@mbiarnes Fix worked well for me locally.
Thanks!

manstis:
Whilst this no doubt fixes the issue; surely `commons-logging:commons-logging` is banned for good reason... perhaps the better fix is to exclude `commons-logging:commons-logging` in the `resteasy-jaxrs` dependency in the applicable `kie-wb-common` module (I forget exactly which it is)?

mareknovotny:
@manstis not sure why do you want to have exclusion down in the build instead of on top level where it makes more sense as this is a ban for all kiegroup build. Am I missing something?

manstis:
@mareknovotny I'm not; it was a proposal that IMO is better than removing the ban. I've been discussing this with @mbiarnes on Friday and we both agree to have the exclusion at the top level.

mareknovotny:
yes, the ban would not be removed. OK then. 
@manstis Are you approving this PR then?

manstis:
@mareknovotny Approved.. the last time I saw this PR it had a change lifting the ban.

I see it's changed now to add the exclusion.

mareknovotny:

PR #1001, opened 2019-07-19
AF-2097: Upgrade JGit library - rollback due to jGit error

paulovmr:
Newer library causes error during initialization of example projects. After investigating and fixing the issue, we will upgrade it again.


paulovmr:

PR #1002, opened 2019-07-19
BAQE-1039 - Change revapi to check against 7.23.0.Final

MarianMacik:
Wait with merge until all valid repositories are good to go.

@rsynek

We now have to check against latest product release (7.4 DM/PAM) which means that we have to check against 7.23.0.Final community release.


MarianMacik:
@rsynek Can you please approve other PRs? I would like to get it merged so it is not desynchronized.

PR #1007, opened 2019-07-23
Further automate revapi cleaning

MarianMacik:
@mbiarnes I have added 2 new options to the git-clone-others.sh file:

* Ability to run it only on a subset of repositories using either `--repo-list=` or `--target-repo=` option. Now it is aligned with other scripts like git-all.sh and mvn-all.sh. *.bat versions lack these parameters, so I decided to not add it to the git-clone-others.bat either.
* I have added `--add-upstream-remote` option so repositories can be cloned and an upstream remote is added (kiegroup) which might be handy in case there is a need to pull the newest upstream changes and rebase on that, for example when creating a PR.

I really have no idea what the `--target-repo=` option is for, so I just wanted to ask.
And please double check it if I didn't break anything. But it should be backward compatible.

### DIFF
--- a/script/git-clone-others.sh
+++ b/script/git-clone-others.sh
@@ -19,6 +19,22 @@ initializeWorkingDirAndScriptDir() {
     # Set script directory and remove other symbolic links (parent directory links)
     scriptDir=`pwd -P`
 }
+
+printUsage() {
+    echo
+    echo "Usage:"
+    echo "  $0 <additional git options> [--repo-list=<list-of-repositories>|--target-repo=<repository>] [--add-upstream-remote]"
+    echo "For example:"
+    echo "  $0"
+    echo "  $0 --bare"
+    echo "  $0 --repo-list=droolsjbpm-knowledge,drools"
+    echo "  $0 --target-repo=drools-wb"
+    echo "In case you are cloning your forks, the following option will create a second remote called upstream"
+    echo "pointing to kiegroup. It can be useful for keeping your forks up to date:"
+    echo "  $0 --add-upstream-remote"
+    echo
+}
+
 initializeWorkingDirAndScriptDir
 droolsjbpmOrganizationDir="$scriptDir/../.."
 
@@ -35,9 +51,42 @@ cd "$droolsjbpmOrganizationDir"
 
 # additinal Git options can be passed simply as params to the script
 # example: --depth 1 (creates a shallow clone with that depth)
-additionalGitOptions="$@"
+additionalGitOptions=()
 
-for repository in `cat "${scriptDir}/repository-list.txt"` ; do
+# default repository list is stored in the repository-list.txt file
+REPOSITORY_LIST=`cat "${scriptDir}/repository-list.txt"`
+
+for arg in "$@"
+do
+    case "$arg" in
+        --target-repo=*)
+            REPOSITORY_LIST=$($scriptDir/checks/repo-dep-tree.pl -w -t ${arg#*=})
+            REPOSITORY_LIST=${REPOSITORY_LIST//,/ }
+        ;;
+
+        --repo-list=*)
+            REPOSITORY_LIST=$(echo "$arg" | sed 's/[-a-zA-Z0-9]*=//')
+            # replace the commas with spaces so that the for loop treats the individual repos as different values
+            REPOSITORY_LIST=${REPOSITORY_LIST//,/ }
+        ;;
+        # add upstream remote for a repository, i.e. git@github.com:kiegroup/${repository}.git, might be useful for PRs
+        --add-upstream-remote)
+            UPSTREAM=true
+        ;;
+
+        --help)
+            printUsage
+            exit 1
+        ;;
+
+        *)
+            additionalGitOptions+=("$arg")
+        ;;
+    esac
+done
+
+
+for repository in $REPOSITORY_LIST ; do
     echo
     if [ -d $repository ] ; then
         echo "==============================================================================="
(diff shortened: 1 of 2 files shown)


script/utils/revapi-clean.sh
```diff
@@ -1,4 +1,19 @@
...
+# ./git-all.sh --repo-list=$REVAPI_REPOSITORIES pull --rebase upstream master
+# ./git-all.sh --repo-list=$REVAPI_REPOSITORIES checkout -b revapi-7.23.0.Final
+# Copy this script to root of all repositories and run it there, ./revapi-clean.sh 7.23.0.Final
+# Update <revapi.oldKieVersion> in kie-parent
+# ./mvn-all.sh --repo-list=$REVAPI_REPOSITORIES clean install -DskipTests
+# ./git-all.sh --repo-list=$REVAPI_REPOSITORIES add -u
+# ./git-all.sh --repo-list=$REVAPI_REPOSITORIES commit -m "BAQE-1039 - Change revapi to check against 7.23.0.Final"
+# ./git-all.sh --repo-list=$REVAPI_REPOSITORIES push origin revapi-7.23.0.Final
+# Create PRs
 # Utility script for removing ignores in revapi-config.json files.
```
mareknovotny: so why all lines are commented if this script helps only as a guide/doc?

script/git-clone-others.sh
```diff
@@ -52,9 +80,17 @@ for repository in `cat "${scriptDir}/repository-list.txt"` ; do
...
+            echo -- additional Git options: ${additionalGitOptions[@]} --
+        fi
+        git clone ${additionalGitOptions[@]} ${gitUrlPrefix}${repository}.git ${repository}
+
+        if [ "$UPSTREAM" = true ]; then
+            upstreamGitUrlPrefix=`echo ${gitUrlPrefix} | sed 's|\(.*github\.com[:/]\).*|\1kiegroup/|'`
+            echo -- adding upstream remote "${upstreamGitUrlPrefix}${repository}.git"
+            cd ${repository}
+            git remote add upstream ${upstreamGitUrlPrefix}${repository}.git
+            cd ..
```
mareknovotny: Isn't by default added origin remote which is in this case the added upstream remote ? Why is this better than the default clone command?

mareknovotny:
@MarianMacik `--target-repo` option is for building all parent repos needed for given target repository
so if you use drools as target repo, it should analyse from inheritance and dependencies what all repositories are needed for that drools repository. That is what I recall and used.

script/git-clone-others.sh
MarianMacik: Well, the first thing you have to do manually is to clone `droolsjbpm-build-bootstrap` so you have this script. After you run it, it will extract the part before `droolsjbpm-build-bootstrap` from the origin remote so this means e.g. in case of my fork `git@github.com:MarianMacik` since origin is my fork, upstream is a kiegroup original repository. But this doesn't matter anyway because all repositories will have only one remote specified, which will be either kiegroup or forked one. But if I want to create revapi PRs, I have to have both the fork and kiegroup remotes specified as I used them to pull the latest changes from upstream and then I commit a new PR on a separate branch which is then pushed to my forks.

So this new option will enable to clone repositories with both remotes specified -> so I can create PRs.

script/utils/revapi-clean.sh
MarianMacik: These are the steps for now (I can make a separate README.md file for revapi if you want) which help to clean revapi ignores semi-automatically. The revapi-clean.sh script is used in one of these steps, specifically on the line 10.

mareknovotny: I meant if it is script you probably want to uncomment these lines to be executed.

MarianMacik: I see your point now. No, these are all steps and the whole file `revapi-clean.sh` is just used for step on the line 10. Otherwise you just copy all commands on command line and execute them one after another.

script/git-clone-others.sh
mareknovotny: ok, I got it as you will clone first your fork for droolsjbpm-build-bootstrap and clone other repos, but then you would add upstream remotes if used that new option. So this changes the order of how it was scripted originally. 
As you cloned all origin repos and add forks only to repos you are contributing.
Could you add this into the README or into the script help?

MarianMacik: well, I can do it also the other way around if you want, but the creation of the second remote wasn't automated until now :)
OK, will update the help.

mareknovotny: I agree, but just to keep clarity and not to confuse someone who already used it the opposite way is I think important.

But I agree your change make it more usable of the script ;)

MarianMacik: `--help` parameter added. I took inspiration from the `mvn-all.sh` script.

script/git-clone-others.sh
```diff
@@ -19,6 +19,21 @@ initializeWorkingDirAndScriptDir() {
...
+    echo
+    echo "Usage:"
+    echo "  $0 <additional git options> [--repo-list=<list-of-repositories>|--target-repo=<repository>] [--add-upstream-remote]"
+    echo "For example:"
+    echo "  $0 --bare"
+    echo "  $0 --repo-list=droolsjbpm-knowledge,drools"
+    echo "  $0 --target-repo=drools-wb"
+    echo "In case you are cloning your forks, the following option will create a second remote called upstream"
+    echo "pointing to kiegroup. It can be useful for keeping your forks up to date:"
+    echo "  $0 --add-upstream-remote"
```
mareknovotny: awesome, thanks for addition!

MarianMacik:
@mbiarnes No, it is just an example. See the first line of help:
```
$0 <additional git options> [--repo-list=<list-of-repositories>|--target-repo=<repository>] [--add-upstream-remote]
```
I took inspiration from here:
https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/622cc7e2f51f6ded866e5341f2da54bf6d45a21c/script/mvn-all.sh#L23-L35

It doesn't mean that I have to call that script like `mvn-all.sh --version`. I have added one more example without any parameters to help function to avoid any confusion. Thanks!

PR #1010, opened 2019-07-25
Move hacep repo after droolsjbpm-integration

winklerm:
This change is needed because openshift-drools-hacep requires kie-maven-plugin which is located in droolsjbpm-integration repository.


winklerm:
@jiripetrlik @mbiarnes could you please take a look?

jiripetrlik:
@winklerm Looks good to me. Thank you!

PR #1025, opened 2019-08-12
RHDM-886: set and override maven-compiler-plugin to version 3.8.1

almope:


pom.xml
```diff
@@ -483,8 +483,9 @@
     <version.org.codehaus.plexus.plexus-io>3.0.0</version.org.codehaus.plexus.plexus-io>
     <version.org.apache.maven.shared.maven-artifact-transfer>0.9.1
     </version.org.apache.maven.shared.maven-artifact-transfer>
-    <version.org.apache.maven.plugins.maven-compiler-plugin>3.7.0
+    <version.org.apache.maven.plugins.maven-compiler-plugin>3.8.1
     </version.org.apache.maven.plugins.maven-compiler-plugin>
+    <version.compiler.plugin>3.8.1</version.compiler.plugin>
```
mareknovotny: duplicate definition?

almope: We need this to override the jboss-parent definition:
https://github.com/jboss/jboss-parent-pom/blob/master/pom.xml#L316

mareknovotny: yes, but we have now 2 version properties in kie-parent, so I would recommend to remove one - `version.org.apache.maven.plugins.maven-compiler-plugin` as that is obviously not used or wrongly used ;)

almope: Updated to remove `version.org.apache.maven.plugins.maven-compiler-plugin`

almope:
Replaced by https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1027

PR #1029, opened 2019-08-15
try to avoid issues with PNC

porcelli:


mareknovotny:
Jenkins execute compile downstream build

mareknovotny:
we obviously need more checking and followed up cleaning if there are other depMgmt here in kie-parent which already are included in individual imported boms.
@mbiarnes could you check them and if there are some remove them?

PR #1056, opened 2019-09-19
New version update

mbiarnes:
uberfire version = kie version
update-version-all.sh has to be fixed and the fixed version in uberfire-bom/pom.xml was replaced by a variable


paulovmr:
I've created and merged a separate PR just to fix the build: https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1057

PR #1057, opened 2019-09-19
Fixing uberfire-bom uberfire version

paulovmr:


paulovmr:
Merging it as master is broken and the fix was tested successfully locally for more than one person.

PR #1070, opened 2019-10-01
BAQE-1114 - Change revapi to check against 7.26.0.Final

MarianMacik:
Wait with merge until all valid repositories are good to go.

@rsynek

We now have to check against latest product release (7.5 DM/PAM) which means that we have to check against 7.26.0.Final community release.


MarianMacik:
@rsynek everything is green now.

PR #1076, opened 2019-10-09
Work around flaky tests by Surefire automatically rerunning them

triceo:


triceo:
@manstis , @mswiderski I think that this could help with the flakiness that we're seeing in some jBPM tests and some UI tests. One thing I am not sure about is whether or not these settings are inherited by your repos or if I need to add it manually there as well.

domhanak:
Jenkins execute full downstream build!

mareknovotny:
retrigerring
Jenkins execute full downstream build!

triceo:

baldimir:
This reduces the possibility of discovering complicated race conditions. I would suggest to revert this here and apply it to the repositories where it is really needed.

PR #1082, opened 2019-10-16
removed obsolete sed commando in script

mbiarnes:


mareknovotny:
I don't really have a clue if that is needed or not ;).

PR #1091, opened 2019-10-31
RHPAM-2466 naraya-integration-bom project created

Ginxo:

PR #1105, opened 2019-11-14
Revert "RHPAM-2466 naraya-integration-bom project created"

yurloc:
Reverts kiegroup/droolsjbpm-build-bootstrap#1091

Jenkins [PR build failed](https://rhba-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/KIE/job/master/job/pullrequest/job/optaweb-employee-rostering-pullrequests/85/console) with error that seems to be related to #1091. Since #1091 wasn't merged after running a full downstream build, it is not certain whether the mentioned build error is caused by #1091 or some other environment instability.

I recommend to merge this ASAP and open a new PR with changes from #1091 and run FDB before it's merged.


Ginxo:
checking https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1091 again before taking a decision

mareknovotny:
@yurloc it seems that our local Nexus mirror is in trouble with running out of space and that error can be related.

yurloc:
@mareknovotny No it's not. I've reproduced the error locally (should've done that before opening this to avoid the uncertainty).

@Ginxo If you can confirm it is a problem it would be great. In that case, I'm 75% convinced that this should be merged immediately because even if there is a quick and obvious fix we still should let it pass through FDB, which is almost 8 hours and we can't wait that long to have the build fixed.

mareknovotny:
ok, @yurloc is right, here is used the dependencies with version properties which is now not declared in kie-parent https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-spring-boot/pom.xml#L74
we need to change/fix the downstream usage of these properties, so reverting here and back then to reopen with fix is preffered

mareknovotny:
@Ginxo please open the PR again with additional PR for https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-spring-boot/pom.xml#L74

yurloc:
Yes this whole block of dependencies needs to be removed: https://github.com/kiegroup/droolsjbpm-integration/blob/7.29.0.Final/kie-spring-boot/pom.xml#L65-L85. Verified locally that it fixes build of the jbpm-spring-boot-sample-basic project.

@Ginxo Note that there might be (low probability but possible) more similar overrides therefore we need a full downstream build for the next attempt.

CC @winklerm - might be interested to see why the build failed even he tested the the narayana-integration-bom usage.

mareknovotny:
fyi @yurloc kiegroup/droolsjbpm-build-bootstrap/pull/1106

PR #1116, opened 2019-11-19
PLANNER-1716 Switch OptaPlanner to junit-vintage-engine

rsynek:
Counterpart in OptaPlanner:
https://github.com/kiegroup/optaplanner/pull/615


ge0ffrey:
@mareknovotny Radovan tried changing everything to junit-vintage (so run with the new runner, but use the old junit 4 api). Unfortunately, the lienzo stuff does some magic, so we limited the scope to only do it for OptaPlanner.

PR #1146, opened 2019-12-17
update to latest ecj

mariofusco:


mareknovotny:
@MarianMacik I am not sure how this is related when the version of ecj was there like 6 months. Could you explain your reference more?

mareknovotny:

mareknovotny:
Is this still related and needed https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-screens/kie-wb-common-project-editor/kie-wb-common-project-editor-client/pom.xml#L205-L210 ?

MarianMacik:
@mareknovotny There has been a change introduced with this PR https://github.com/kiegroup/drools/pull/2650 which translated Java 10+ option to something ECJ wasn't able to accept, so it used default, which is Java 1.4 :) I made a quick fix in https://github.com/kiegroup/drools/pull/2693 but @mariofusco decided that we should upgrade it right away since there is already a version available which supports up to Java 13 I believe.

mareknovotny:
are we sure that FDB is/was green?

PR #1147, opened 2019-12-17
[PLANNER-1753] Bump and add Spring Boot dependency management for optaplanner-integration

cuijulian:
Jira: https://issues.redhat.com/browse/PLANNER-1753

Prerequisite for: https://github.com/kiegroup/optaplanner/pull/640

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -306,8 +306,8 @@
     <version.org.sonatype.plexus.plexus-sec-dispatcher>1.3</version.org.sonatype.plexus.plexus-sec-dispatcher>
     <version.org.sonatype.sisu>2.3.0</version.org.sonatype.sisu>
     <version.org.sonatype.sisu.sisu-guice>3.2.3</version.org.sonatype.sisu.sisu-guice>
-    <version.org.springframework>5.2.1.RELEASE</version.org.springframework>
-    <version.org.springframework.boot>2.2.1.RELEASE</version.org.springframework.boot>
+    <version.org.springframework>5.2.2.RELEASE</version.org.springframework>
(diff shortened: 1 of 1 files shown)


cuijulian:

cuijulian:

pom.xml
```diff
@@ -5151,6 +5151,32 @@
...
+      <dependency>
+        <groupId>org.springframework.boot</groupId>
+        <artifactId>spring-boot-test</artifactId>
+        <version>${version.org.springframework.boot}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.springframework.boot</groupId>
+        <artifactId>spring-boot-test-autoconfigure</artifactId>
+        <version>${version.org.springframework.boot}</version>
+      </dependency>
```
yurloc: I think we shouldn't replicate dependency management of `spring-boot-dependencies`.

ge0ffrey: Why stop now?  (just kidding)
I agree, we should into reusing their bom, if and only if it only contains spring dependency. So not reuse their platform bom (which also includes hibernate dependencies etc).

But for me, that doesn't block this PR as this one just continues the way we've been doing it so far.

@cuijulian Can you make a jira for trying to reuse the spring bom (the one without hibernate deps etc, if such one exists)?

cuijulian: @ge0ffrey The `spring-boot-dependencies` POM definitely manages hibernate dependencies. There should be a way to exclude them when importing. But there may be a Spring BOM without them, I'll have a look. I'll make a Jira, this will be in a separate PR.

PR #1186, opened 2020-02-12
BAQE-1054 Enable cross-module coverage for XML reports

rsynek:
This change returns back a single jacoco.exec file. For XML report generation this approach leverages jacoco ant goal "report", as jacoco-maven-plugin does not provide enough control of scope of the report.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1955,6 +1955,18 @@
                   <goal>run</goal>
                 </goals>
                 <configuration>
+                  <!--
+                    Jacoco ant "report" task provides control over scope of the generated report. The report task
+                    requires access to sources, classes and .exec file containing coverage data. The configuration
+                    below uses sources and classes of the entire project (each of its modules) and a single jacoco.exec
+                    file placed in project root directory.
+                    Jacoco maven plugin does not provide such a level of control and requires an artificial module that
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -1944,20 +1946,53 @@
       <build>
         <plugins>
           <plugin>
-            <groupId>org.jacoco</groupId>
-            <artifactId>jacoco-maven-plugin</artifactId>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-antrun-plugin</artifactId>
```
triceo: I think that this deserves some documentation, at least in the form of an explanatory XML comment.

mareknovotny: @triceo what kind of documentation?  I would be fine with adding here some comments and show usage in [README.md](https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/README.md)

triceo: Radek already added an XML comment, I consider that good enough.

mareknovotny: ok

rsynek:
jenkins run compile downstream build

rsynek:
@mareknovotny PR checks passed; would you please merge it?

PR #1199, opened 2020-02-27
Added WebJar for Swagger UI dependency

gmunozfe:
Needed for Swagger-UI tests with Springboot, following [documentation](https://github.com/kiegroup/kie-docs/blob/master/doc-content/enterprise-only/springboot/bus-app-swagger_proc.adoc)


sutaakar:
@gmunozfe Please change the branch to be same as https://github.com/kiegroup/droolsjbpm-integration/pull/2030 so Jenkins can use both PRs at the same time during the build.

PR #1223, opened 2020-03-09
[BAQE-1323] DB scripts move to a new module along with refactoring work in jbpm-installer

afalhambra:
- A new jbpm-db-scripts module has been included to hold the db scripts for jbpm and quartz.


MarianMacik:
@gmunozfe can you approve as well?

PR #1225, opened 2020-03-11
KOGITO-1107: Ensure prod daily build pass kie-wb-common selenium test

jomarko:
Ensemble together with:
- https://github.com/kiegroup/kie-wb-common/pull/3217
- https://github.com/kiegroup/kie-jenkins-scripts/pull/702

### DIFF
--- a/script/mvn-all.sh
+++ b/script/mvn-all.sh
@@ -40,7 +40,7 @@ droolsjbpmOrganizationDir="$scriptDir/../.."
 # default repository list is stored in the repository-list.txt file
 REPOSITORY_LIST=`cat "${scriptDir}/repository-list.txt"`
 
-# default repository list where it tests are skipped is stored in the repository-list.txt file
+# default repository list where integration tests are skipped is stored in the skip-it-repository-list.txt file
 SKIP_IT_REPOSITORY_LIST=`cat "${scriptDir}/skip-it-repository-list.txt"`
 
 MVN_ARG_LINE=()
--- a/script/mvn-all.sh
+++ b/script/mvn-all.sh
@@ -39,6 +39,10 @@ droolsjbpmOrganizationDir="$scriptDir/../.."
 
 # default repository list is stored in the repository-list.txt file
 REPOSITORY_LIST=`cat "${scriptDir}/repository-list.txt"`
+
+# default repository list where it tests are skipped is stored in the repository-list.txt file
+SKIP_IT_REPOSITORY_LIST=`cat "${scriptDir}/skip-it-repository-list.txt"`
+
 MVN_ARG_LINE=()
 
 for arg in "$@"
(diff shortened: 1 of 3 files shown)


script/mvn-all.sh
```diff
@@ -39,6 +39,10 @@ droolsjbpmOrganizationDir="$scriptDir/../.."
 
 # default repository list is stored in the repository-list.txt file
 REPOSITORY_LIST=`cat "${scriptDir}/repository-list.txt"`
+
+# default repository list where it tests are skipped is stored in the repository-list.txt file
```
mbiarnes: @jomarko is "it" too much or its or the

jomarko: fixed, thank you

mareknovotny:
do we need to introduce new list of repos? I mean if we disable that in default reactor build it should be enough, or not?

jomarko:
sorry, I am not sure what is 'default reactor build'

jomarko:
I need a WIN expert help to update also `mvn-all.bat`

mareknovotny:
> sorry, I am not sure what is 'default reactor build'

I mean default build flow without specifying profile or property which turns on the special profile ;)

jomarko:
@mareknovotny sorry, I am still not sure what is your idea about the fix for this, I will summarize my findings here maybe it will help:
- we use two different groovy files for job specification **pr_jobs.groovy** and **dailyBuild_pipeline.groovy**
- the problem is they have different granularity/approach for starting builds of repositories
- **pr_jobs.groovy** is written in way that started job has configuration according to a repository, where pull request was opened, so it is easy to decide what labels the slave needs, what maven profiles needs to be activated, what system property need to be passed
- **dailyBuild_pipeline.groovy** is written in way that all repositories are started by one generic command **05a_communityDeployLocally.sh**  and all are started on the same slave, that has no **gui-testing** label, what is logical, because total count of selenium tests is pretty small and we need that label just for two repositories: **kie-wb-common** and **kie-wb-distributions** 
- The selenium tests in **kie-wb-common** are integration tests included in **kogito** profile, that is activated, just if **-Dproductized** is missing, so if we check **05a_communityDeployLocally.sh** and **mvn-all.sh** it is activated in daily build pipeline, however at that moment slave has no **gui-testing** label, so we need so skip the tests
- Then in **dailyBuild_pipeline.groovy** we have a separate job for selenium tests - with **gui-testing** label so I try to start the new selenium tests in this moment. 

If you have specific concept in mind, feel free to open PR against my changes.

mareknovotny:
my main issue is with creating new file `script/skip-it-repository-list.txt` and all logic in mvn-all scripts which needs to be maintained while if you skip selenium tests by default it would be the same. 
If you need to run selenium tests just activate that through -Dproperty or by additional profile. That would be easier to get or am I missing something here?

jomarko:
@mareknovotny ok so you ask me to fix without changes in this repository

mareknovotny:
you could change script(s) if there is a need too, but I would not recommend to add that logic around  `script/skip-it-repository-list.txt`

PR #1230, opened 2020-03-16
BXMSPROD-583 cleanWS removed and DEFAULT_BRANCH job parameter adapted

Ginxo:
BXMSPROD-583

### DIFF
--- a/Jenkinsfile.nightly
+++ b/Jenkinsfile.nightly
@@ -16,8 +16,6 @@ pipeline {
         stage ("Clean") {
             steps {
                 echo "[INFO] Cleaning the workspace."
-                cleanWs()
-                echo "[SUCCESS] Workspace was successfully cleaned up."
             }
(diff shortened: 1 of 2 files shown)


Jenkinsfile.nightly
```diff
@@ -16,8 +16,6 @@ pipeline {
         stage ("Clean") {
             steps {
                 echo "[INFO] Cleaning the workspace."
```
winklerm: Wondering if this echo should be there?

mareknovotny: well, the whole stage Clean is then useless, why not remove it at all?

Ginxo: removed

Ginxo:
Since we have transformed the multibranch pipeline to pipeline with parameters because of QE requirements the pipeline should be adapted.

PR #1231, opened 2020-03-16
BXMSPROD-583 nightly pipeline file created for 7.33.x

Ginxo:
BXMSPROD-583


Ginxo:
@winklerm the seetings file id is 5d9884a1-178a-4d67-a3ac-9735d2df2cef is an specific one for nightly. Can you please take a look and check if it's right or not?

PR #1236, opened 2020-03-16
BXMSPROD-583 Jenkinsfile with parameters

Ginxo:
BXMSPROD-583


mareknovotny:
we need to change also the image label(s)

Ginxo:
@mareknovotny @winklerm `kie-rhel7 && kie-mem24g` used instead

PR #1250, opened 2020-03-25
BXMSPROD-710 upstream/downstream pipelines adatped to github plugin requirements

Ginxo:
BXMSPROD-710
kie-mem24g label is needed since we use the same pipeline for community and non-community projects


Ginxo:
kie-mem24g label is needed since we use the same pipeline for community and non-community projects

PR #1254, opened 2020-03-27
BXMSPROD-719 nightly tweak backported to 7.33.x

Ginxo:
BXMSPROD-719

### DIFF
--- a/Jenkinsfile.nightly
+++ b/Jenkinsfile.nightly
@@ -11,7 +11,7 @@ pipeline {
     parameters {
         string(description: 'The deployment URL', name: 'KIE_GROUP_DEPLOYMENT_REPO_URL')
         booleanParam(description: 'Skip Tests? True as default', name: 'SKIP_TESTS', defaultValue: true)
-        string(description: 'The UMB message version', name: 'UMB_VERSION', defaultValue: 'master')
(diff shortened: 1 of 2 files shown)


Jenkinsfile.nightly
```diff
@@ -9,8 +9,10 @@ pipeline {
         jdk 'kie-jdk1.8'
     }
     parameters {
-        string(defaultValue: 'master', description: 'The default branch', name: 'DEFAULT_BRANCH')
         string(description: 'The deployment URL', name: 'KIE_GROUP_DEPLOYMENT_REPO_URL')
+        booleanParam(description: 'Skip Tests? True as default', name: 'SKIP_TESTS', defaultValue: true)
+        string(description: 'The UMB message version', name: 'UMB_VERSION', defaultValue: 'master')
```
winklerm: Should the defaultValue `master` be here? I know it is overridden when the pipeline is actually run, but maybe it would be safer to put here a default for `7.33.x`?

Ginxo: In case we say 7.33.x we will have to update this file each time the default branch is deprecated. I propose to remove the default in case master is not right.

Ginxo: @winklerm I just noticed this is the backported one. So fixing it.

PR #1256, opened 2020-03-28
[BXMSPROD-709]: introduced jackson-bom with special version property

mareknovotny:
https://issues.redhat.com/browse/BXMSPROD-709

merge together with kiegroup/droolsjbpm-integration/pull/2068

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -74,8 +74,10 @@
     </version.com.ahome-it.lienzo.tooling.nativetools>
     <version.com.ahome-it.lienzo.tooling.common>1.0.190-RELEASE</version.com.ahome-it.lienzo.tooling.common>
     <version.com.allen-sauer.gwt.dnd>3.3.3</version.com.allen-sauer.gwt.dnd>
-    <version.com.fasterxml.jackson>2.9.9</version.com.fasterxml.jackson>
-    <version.com.fasterxml.jackson.databind>2.9.10.3</version.com.fasterxml.jackson.databind>
+    <!-- keep these 2 version properties for jackson to have working replacement for camel/karaf build -->
+    <version.com.fasterxml.jackson>2.9.10</version.com.fasterxml.jackson>
+    <version.com.fasterxml.jackson.databind>2.9.10.4</version.com.fasterxml.jackson.databind>
+    <version.com.fasterxml.jackson.bom>2.9.10.20200411</version.com.fasterxml.jackson.bom>
     <version.com.google.code.gson>2.8.2</version.com.google.code.gson>
     <version.com.google.guava>25.0-jre</version.com.google.guava>
     <version.com.google.inject.guice>4.0</version.com.google.inject.guice>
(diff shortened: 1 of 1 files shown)


mareknovotny:

mareknovotny:
@yurloc I tried to use jackson-bom and changed version properties - declare new one for jackson-bom and removed our jackson version properties, but there are places in droolsjbpm-integration repository like https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-osgi/kie-karaf-features/src/main/filtered-resources/repository/features.xml#L80 and https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-osgi/kie-karaf-features/src/main/filtered-resources/repository/features-fuse.xml#L117 which requires to have version properties declared.

mareknovotny:

mareknovotny:
it looks good even 2 failures which are in daily build now too
```
    org.jbpm.test.functional.task.HumanTaskQueryFilterTest.testSingleResult
    org.jbpm.test.functional.task.HumanTaskQueryFilterTest.testFilterParams
```
filed jiras for these failures as https://issues.redhat.com/browse/JBPM-9085 and https://issues.redhat.com/browse/JBPM-9086

mareknovotny:

pom.xml
```diff
@@ -74,7 +74,8 @@
     </version.com.ahome-it.lienzo.tooling.nativetools>
     <version.com.ahome-it.lienzo.tooling.common>1.0.190-RELEASE</version.com.ahome-it.lienzo.tooling.common>
     <version.com.allen-sauer.gwt.dnd>3.3.3</version.com.allen-sauer.gwt.dnd>
-    <version.com.fasterxml.jackson>2.9.9</version.com.fasterxml.jackson>
+    <version.com.fasterxml.jackson.bom>2.9.10.20200223</version.com.fasterxml.jackson.bom>
+    <version.com.fasterxml.jackson>2.9.10</version.com.fasterxml.jackson>
```
yurloc, line 78: I would recommend to add a comment here explaining where this property is used and why it must not be removed. It appears to be unused so without the explanation, one could remove it mistakenly or unnecessarily spend time looking for the usage.

mareknovotny:

mareknovotny:

mareknovotny:

mareknovotny:

mareknovotny:
@yurloc there are also 3 failures from employeerostering repository
```
org.optaweb.employeerostering.rotation.RotationServiceTest.createShiftTemplateTest
org.optaweb.employeerostering.rotation.RotationServiceTest.updateShiftTemplateTest
org.optaweb.employeerostering.rotation.RotationServiceTest.getShiftTemplateTest
```
are these legitimate failures?

yurloc:
@mareknovotny Yes those are test failures caused by the upgrade. I'd like to ask @Christopher-Chianelli to fix the tests but I'm not sure whether it's possible to build PR from different people together.

If that's not possible, I suggest that you (Marek) open a PR to  employee rostering that temporarily downgrades Jackson. That will allow you to merge this with a green FDB and us to re-align independently.

Christopher-Chianelli:
Issue is probably caused by https://github.com/FasterXML/jackson-modules-java8/pull/75 (Duration are now written as a timestamp instead of a ISO duration string). Might be able to revert it to previous behavior by configuring the ObjectMapper.

mareknovotny:
ok, thanks @yurloc @Christopher-Chianelli for quick feedback, I am split into more things but I will try to fix it and let review it.
I am only confused why it is shwing in not minor upgrade and the answer could be it is leaking different version of jackson into your dependencies like it is in spring-boot. I will look at that leakage to properly manage 2.9.10.4 not the 2.10 ;)

yurloc:
@mareknovotny I think it could be because optawebs import `spring-boot-dependencies` BOM and that overrides dependency management from `kie-parent`.

mareknovotny:
@yurloc @Christopher-Chianelli optaweb-employee-rostering uses spring-boot-dependencies BOM which even before this PR use jackson 2.10.1, so the failures are not related to this change where I try to introduce the jackson BOM 2.9.10. We need to avoid usage of spring-boot managed version of jackson or let it be here in this PR.

PR #1261, opened 2020-04-02
BXMSPROD-710 skipTests treatment

Ginxo:
BXMSPROD-710
Depends on https://github.com/kiegroup/jenkins-pipeline-shared-libraries/pull/18

### DIFF
--- a/downstream.production.stages
+++ b/downstream.production.stages
@@ -14,6 +14,6 @@ stage('Downstream Production Build') {
 
         println "File ${REPOSITORY_LIST_FILE} and ${PRODUCTION_PROJECT_LIST} jenkins file merged in ${projectCollection}"
 
-        treebuild.downstreamBuild(projectCollection, "${SETTINGS_XML_ID}", 'clean install -Dproduction', false)
+        treebuild.downstreamBuild(projectCollection, "${SETTINGS_XML_ID}", 'clean install -Dproduction')
     }
 }
--- a/downstream.production.stages
+++ b/downstream.production.stages
@@ -14,6 +14,6 @@ stage('Downstream Production Build') {
 
         println "File ${REPOSITORY_LIST_FILE} and ${PRODUCTION_PROJECT_LIST} jenkins file merged in ${projectCollection}"
 
-        treebuild.downstreamBuild(projectCollection, "${SETTINGS_XML_ID}", 'clean install -Dproduction')
+        treebuild.downstreamBuild(projectCollection, "${SETTINGS_XML_ID}", 'clean install -Dproductized=true')
     }
 }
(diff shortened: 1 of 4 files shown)


downstream.production.stages
```diff
@@ -14,6 +14,6 @@ stage('Downstream Production Build') {
 
         println "File ${REPOSITORY_LIST_FILE} and ${PRODUCTION_PROJECT_LIST} jenkins file merged in ${projectCollection}"
 
-        treebuild.downstreamBuild(projectCollection, "${SETTINGS_XML_ID}", 'clean install -Dproduction', false)
+        treebuild.downstreamBuild(projectCollection, "${SETTINGS_XML_ID}", 'clean install -Dproductized=true')
```
mareknovotny, line 17: good catch

upstream.stages
```diff
@@ -8,6 +8,6 @@ stage('Upstream Build') {
         def gitURL = env.ghprbAuthorRepoGitUrl ?: env.GIT_URL
         def project = treebuild.getProjectGroupName(treebuild.getProject(gitURL))[1]
 
-        treebuild.upstreamBuild(projectCollection, project, "${SETTINGS_XML_ID}", 'clean install', false)
+        treebuild.upstreamBuild(projectCollection, project, "${SETTINGS_XML_ID}", '-e -nsu -fae -Pbusiness-central,wildfly,sourcemaps,no-showcase clean install -Dproduction -Dfull=true -Dcontainer=wildfly -Dcontainer.profile=wildfly -Dintegration-tests=true -Dmaven.test.failure.ignore=true -Dmaven.test.redirectTestOutputToFile=true -Dgwt.compiler.localWorkers=1')
```
mareknovotny, line 11: Is this correct? 
I think original upstream maven command was simple just maven clean install for getting the bits built
see https://github.com/kiegroup/kie-jenkins-scripts/blob/master/job-dsls/jobs/pr_jobs.groovy#L12 line and the command from it here:
```
"-B -e -T1C -s \$SETTINGS_XML_FILE -Dkie.maven.settings.custom=\$SETTINGS_XML_FILE -DskipTests -Dgwt.compiler.skip=true -Dgwt.skipCompilation=true -Denforcer.skip=true -Dcheckstyle.skip=true -Dspotbugs.skip=true -Drevapi.skip=true clean install
```
The maven command here is for downstream builds

Ginxo: ok, reverting it...

mareknovotny: but are the enforcer and some others properties applied with just short `mvn clean install` it should not be running such plugin execution ;)

Ginxo: @mareknovotny the final mvn command is defined here 
https://github.com/kiegroup/jenkins-pipeline-shared-libraries/blob/f49ed111275fc503ae3cf03d0027efcf098a3426/vars/maven.groovy#L11
so `mvn -B -s $MAVEN_SETTINGS_XML -fae clean install` will be executed. Is it right?

mareknovotny: honestly not sure as I can't see these properties/options which takes executions not wanted in upstream builds
`-DskipTests -Dgwt.compiler.skip=true -Dgwt.skipCompilation=true -Denforcer.skip=true -Dcheckstyle.skip=true -Dspotbugs.skip=true -Drevapi.skip=true`

PR #1263, opened 2020-04-06
PLANNER-1912: Move Pitest config so that it can be overriden easily

yurloc:
merge with this kiegroup/optaweb-vehicle-routing/pull/290


mareknovotny:

PR #1276, opened 2020-04-21
BXMSPROD-575 Jenkinsfile.nightly notification emails

Ginxo:
https://issues.redhat.com/browse/BXMSPROD-575

PR #1281, opened 2020-04-24
BXMSPROD-575 success post action email added

Ginxo:
https://issues.redhat.com/browse/BXMSPROD-575
Related with https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1276#pullrequestreview-397110852
I guess the success is also important to be notified, isn't it?

### DIFF
--- a/Jenkinsfile.nightly
+++ b/Jenkinsfile.nightly
@@ -125,17 +125,23 @@ pipeline {
         }
     }
     post {
-        unstable {
-            script {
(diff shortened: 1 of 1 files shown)


mareknovotny:
@Ginxo I am not 100% sure we need that as in success status the pipeline sends UMB message to wired other pipelines, so it is at the end ok to not send another duplicate email notification. But I am not strongly against this ;)

Ginxo:
Closed since UMB message should be enough. We will re-create it if it's needed

PR #1284, opened 2020-04-27
BXMSPROD-736: upgdating README.md

mbiarnes:
the passages if Developing with Eclipse and Developing with IntelliJ are not yet updated becuse I think best is that this someone who uses this does this - I am not the best person to do this. There are several links to guvnor (thta is now drools-wb) and GWT configuration.
A general question: should these be maintained? I mean likewise no one uses this and this is quit hard to maintain.


README.md
```diff
@@ -1276,51 +1280,15 @@ Knowing what's going on
...
-            * Go to the Jenkins job of the projects you're working on:
-
-                * [guvnor](https://hudson.jboss.org/hudson/view/Drools%20jBPM/job/guvnor/)
-
-            * Right click in the lower left corner on the *All* feed link, menu item *Add link to Jenkins build monitor*.
-
-        * Otherwise, check [the Jenkins website](https://hudson.jboss.org/hudson/view/Drools%20jBPM/) often.
-
-            * Note: the public Jenkins is a mirror of the VPN internal Red Hat Jenkins and is sometimes stale.
+        * <kiegroup repository> i.e. https://github.com/kiegroup/<repository>/commits/master.atom
```
mareknovotny, line 1287: as an example this should be exact URL to check like https://github.com/kiegroup/droolsjbpm-build-bootstrap/commits/master.atom

yesamer:
> AFAIK the Eclipse section can be removed; IDK anybody who would be able to update it.
> 
> @yesamer would you be so kind as to help @mbiarnes review/update the InteliJ instructions to run `drools-wb`?

@manstis With great pleasure

README.md
mbiarnes: @mareknovotny  Look at the example above - this is the droolsjbpm-build-bootstrap example. The line starting with <kiegroup repository> should be an example how to construct the URL. You see the line below - it links you to the file which reps we have in our kiegroup.
Maybe is is not clear that this is like an example. I didn't want to repeat 25 the same - only changing the rep name.

mbiarnes: @mareknovotny 
Better explained now

mareknovotny:
> AFAIK the Eclipse section can be removed; IDK anybody who would be able to update it.
> 
> @yesamer would you be so kind as to help @mbiarnes review/update the InteliJ instructions to run `drools-wb`?

I don't think we should prefer one IDE before another. We would remove/move the IDE setup somewhere else probably as individual repositories needs different plugins and setup anyway

mbiarnes:
@mareknovotny @yesamer 
to remove the InteliJ passage would be the most easy and the fastest ;)

yesamer:
> @mareknovotny @yesamer
> to remove the InteliJ passage would be the most easy and the fastest ;)

@mareknovotny Therefore the required updated instructions are no longer necessary, is that right?

mareknovotny:
I think we should add IDE specific setup to individual repositories if they are really needed, here it is not ok I think

yesamer:
> I think we should add IDE specific setup to individual repositories if they are really needed, here it is not ok I think

Well, I don't have any evidence/reason that `drools-wb` can works only on IntelliJ, even if I never tried alternative IDEs. Considering this, I think it's not necessary. Agree with you to proceed without it.

mareknovotny:
just to not confuse you with my previous comment, the README.md tries to show the basic setup and the drools-wb was chosen as a replacement of guvnor where it was an example setup done. There can be any project shown in the example setup which use GWT and kie-wb-* components

yesamer:
@mareknovotny If I correctly understood, we just need to put an example here to run a GWT project, and `drools-wb` was chosen for it. 
In this case:

I analyzed the current description present here https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1284/files#diff-04c6e90faac2675aa89e2176d2eec7d8R864. 
I suggest to dismiss usage of `GWT Configuration` extension at all, and to substitute it with using a maven command ( `mvn gwt:run` ) 
In this way, we will have the follow advantages:
- Simpler instructions;
- Instructions will apply for any IDE (It will work even without an IDE);
- No need to update instructions.
- All required parameters and setting are managed inside `drools-wb-webapp\pom.xml` file (https://github.com/kiegroup/drools-wb/blob/master/drools-wb-webapp/pom.xml#L1573)

To more clear, in case of using maven command, the steps I suggest are similar to:
**To launch drools-wb in GWT SuperDev mode:**
1. Move into `drools-wb/drools-wb-webapp-directory`
2. Launch `mvn clean process-resources gwt:run` 
2a. Launch `mvn clean process-resources gwt:debug` for launch it in debug mode

In case of `GWT Configuration` extension usage:
**Add new GWT configuration**

Combobox Module: `drools-wb-webabb`

Combobox GWT Module to load: `org.drools.workbench.DroolsWorkbench`

Textfield VM options: `-Xmx8G -Xms1024m -Xss1M -XX:CompileThreshold=7000 -Derrai.jboss.home="/target/wildfly-14.0.1.Final"  ....... `

WDYT? @mbiarnes @mareknovotny @manstis

yesamer:
My only doubt is: have all our GWT projects configured the `gwt-maven-plugin` in the same way?

mareknovotny:
Can you answer the https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1284#issuecomment-621299451 question @yurloc, @ederign, @romartin @bxf12315 please?

bxf12315:
> My only doubt is: have all our GWT projects configured the `gwt-maven-plugin` in the same way?

jbpm-wb side use same plugin, https://github.com/kiegroup/jbpm-wb/blob/ea65314c6b244c7ec443090ee40a6d1c5b358ec6/jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-showcase/pom.xml#L525 and https://github.com/kiegroup/jbpm-wb/blob/ea65314c6b244c7ec443090ee40a6d1c5b358ec6/jbpm-wb-showcase/pom.xml#L1761

romartin:
as far as I know, at least fo the `kie-wb-common` repository, we're using the "regular" gwt plugin

PR #1325, opened 2020-05-22
Update optaplanner quickstarts version

ge0ffrey:
Depends on https://github.com/kiegroup/optaplanner/pull/783

### DIFF
--- a/script/release/update-version-all.sh
+++ b/script/release/update-version-all.sh
@@ -143,6 +143,15 @@ for repository in `cat ${scriptDir}/../repository-list.txt` ; do
             sed -i "s/<version.org.kie>$kieOldVersion<\/version.org.kie>/<version.org.kie>$newVersion<\/version.org.kie>/" pom.xml
             returnCode=$?
 
+        elif [ "$repository" == "optaplanner" ]; then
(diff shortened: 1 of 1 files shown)


script/release/update-version-all.sh
```diff
@@ -143,6 +143,15 @@ for repository in `cat ${scriptDir}/../repository-list.txt` ; do
             sed -i "s/<version.org.kie>$kieOldVersion<\/version.org.kie>/<version.org.kie>$newVersion<\/version.org.kie>/" pom.xml
             returnCode=$?
 
+        elif [ "$repository" == "optaplanner" ]; then
+            mvnVersionsUpdateParentAndChildModules
+            returnCode=$?
+            if [ $returnCode != 0 ] ; then
+                exit $returnCode
+            fi
+            optaplanner-quickstarts/update-version.sh $newVersion $releaseType
```
mbiarnes, line 152: ... or optaplanner-quickstarts/update-version.sh $newVersion community
since productized is not used any more! Both cases should work. So what you like more.

ge0ffrey: Let's just pass it along, for consistency if you do ever end up using it again?

In any case, the other script currently doesn't do anything with it.

PR #1327, opened 2020-05-25
PLANNER-1938 Add dependency mgmt for com.sun.activation

rsynek:


mareknovotny:
shouldn't it be jakarta instead of javax?
```
        <dependency>
            <groupId>com.sun.activation</groupId>
            <artifactId>jakarta.activation</artifactId>
```

rsynek:
I will open a new PR, as there is going to be a broader change of dependencies in kie-server tests using a different branch. This PR will be closed.

PR #1233, opened 2020-03-16
BXMSPROD-685: upgraded version.org.ow2.asm

mbiarnes:

PR #1337, opened 2020-05-27
Upgrade Plexus, Maven, sisu, httpclient, guava, guice, cli, codec, asm, jupiter and jandex

ge0ffrey:
Upgrades a bunch of components that haven't been updated in ages.
Experiment to see how much this breaks the downstream build :)

Open question:
- Are any of these components defined by WildFly/EAP? Which version do they use?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -54,12 +54,12 @@
       - Version properties must be sorted alphabetically (other form of sorting were found to be unclear and ambiguous).
     -->
     <version.aopalliance>1.0</version.aopalliance>
-    <version.asm>3.3.1</version.asm>
+    <version.asm>7.3.1</version.asm>
     <version.ch.qos.cal10n>0.8.1</version.ch.qos.cal10n>
     <version.ch.qos.logback>1.2.3</version.ch.qos.logback>
     <version.commons-beanutils>1.9.4</version.commons-beanutils>
-    <version.commons-cli>1.3.1</version.commons-cli>
-    <version.commons-codec>1.11</version.commons-codec>
+    <version.commons-cli>1.4</version.commons-cli>
+    <version.commons-codec>1.13</version.commons-codec>
     <version.commons-collections>3.2.2</version.commons-collections>
     <version.commons-configuration>1.6</version.commons-configuration>
     <version.commons-dbcp>1.4</version.commons-dbcp>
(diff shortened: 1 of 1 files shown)


ge0ffrey:

pom.xml
```diff
@@ -261,7 +261,7 @@
     <version.org.wildfly>18.0.1.Final</version.org.wildfly>
     <version.org.wildfly.arquillian>2.2.0.Final</version.org.wildfly.arquillian>
     <version.org.wildfly.core>10.0.3.Final</version.org.wildfly.core>
-    <version.org.wildfly.common>1.5.1.Final</version.org.wildfly.common>
+    <version.org.wildfly.common>1.5.4.Final-format-001</version.org.wildfly.common>
```
mareknovotny, line 264: this could be controversial upgrade due the version is that really valid thing due our versioning here in JBoss MW?

pom.xml
```diff
@@ -318,7 +318,7 @@
     <version.org.jboss.errai.wildfly>${version.org.wildfly}</version.org.jboss.errai.wildfly>
     <!-- Required by jboss-remoting version above -->
     <version.org.wildfly.security>1.10.4.Final</version.org.wildfly.security>
-    <version.org.ow2.asm>5.0.4</version.org.ow2.asm>
+    <version.org.ow2.asm>8.0.1</version.org.ow2.asm>
```
mareknovotny, line 321: we already are trying to upgrade to 7.1 at https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1233.

pom.xml
ge0ffrey: Good point. Shall I change it to 1.5.4.Final?

pom.xml
ge0ffrey: I think 7.1 would probably suffice already too, so once #1233 is merged, I am happy to rebase and remove this line.

mareknovotny: ok

mareknovotny:
so upgraded maven and plexus deps sems are not ok for kie-maven-plugin even maven 3.3.9 is not used anymore in CI it is somehow expected to have it in tests for droolsjbpm-integration kie-maven-plugin?

ge0ffrey:
I am going to withdraw this PR, due to lack of bandwidth to spend on this, it feels too much akin to _solving world hunger._ I've cleaned up parts of kie-parent dependencies trees in the past, adjusting code in drools, jbpm, optaplanner, tycho, osgi tests or other dark corners. It always took multiple weeks and a lot of pain. A burned child dreads the fire. These days few people are left that upgrade anything serious in kie-parent. Big praise to heroes such as MichaelB (hibernate 5.3, ip-bom removal), Marek and Julian Cui (Spring 5) recently. 

That being said, I believe the current set up isn't optimal. Yes, **we must not have every module declare it's own dependency version** - that's impossible to sync before a release (including a product release). **But this one-dependencies-list-to-rule-them-all-pom is the other end of the spectrum: it is unmaintainable**, because 1) any contributor is forced to take responsibility over all projects (= build Rome in a day) and 2) transitive dependency version get overwritten by default (instead of by exception). This results in 1) most our dependencies are outdated and 2) our transitive dependencies are messed up.

I am starting to become a believer in the quarkus/kogito approach: one-dependencies-list-per-project-parent-pom. Then 1) any contributor can update their project in isolation (=> much easier to upgrades dependencies) and 2) transitive dependency overwrites can be minimized (=> much easier to upgrade dependencies). But this comes at the cost of a sync dependency version list across project operation during releases. Worth it, I think.

mareknovotny:
the environment is more prepared for somehing which is more between top level depMgmt and project only depMgmt. We need to look at that, but investing more time for this on v7 will need to see some long profit too.

PR #1351, opened 2020-06-04
The enforcer plugin shouldn't trigger banDuplicatedClasses in quarkus-ide-launcher

ge0ffrey:
That dependency does not appear in product builds, so it has zero impact on product.
In the quarkus extensions, it's a dependency of quarkus-core that gets removed later on magically.
It's a blocker for the quarkus-optaplanner externalization.


ge0ffrey:
execute full downstream build

PR #1355, opened 2020-06-08
[JBPM-9138] - RESTWorkItemHandler doesn't support proxy

fjtirado:
Adding mock server dependency for testing proxy


pom.xml
```diff
@@ -5249,6 +5249,18 @@
         <artifactId>base</artifactId>
         <version>${version.com.google.jsinterop.base}</version>
       </dependency>
+
+      <dependency>
+        <groupId>org.mock-server</groupId>
+        <artifactId>mockserver-netty</artifactId>
+        <version>5.10.0</version>
```
gmunozfe, line 5256: Convert version to a variable property

fjtirado: done

PR #1374, opened 2020-06-24
[DROOLS-5451] Upgraded org.jpmml:pmml-model to 1.5.1 version

gitgabrio:
@tarilabs @jiripetrlik @manstis 

See https://issues.redhat.com/browse/DROOLS-5451

I launch a FDB to verify it does not broke something

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -228,7 +228,7 @@
     <version.org.jboss.xnio>3.7.7.Final</version.org.jboss.xnio>
     <version.org.jdom>1.1.3</version.org.jdom>
     <version.org.jfree.jfreechart>1.5.0</version.org.jfree.jfreechart>
-    <version.org.jpmml.model>1.5.1</version.org.jpmml.model> <!-- jpmml-model BSD 3C license - ATTENTION 1.4.11 intentional, because 1.4.9 evaluators depend on 1.4.11 -->
+    <version.org.jpmml.model>1.5.1</version.org.jpmml.model> <!-- jpmml-model BSD 3C license - ATTENTION 1.5.1 intentional, because 1.5.1 evaluators works with 1.5.1 -->
     <version.org.json>20090211</version.org.json>
     <version.org.jsoup>1.8.3</version.org.jsoup>
     <version.org.junit>5.5.2</version.org.junit>
--- a/pom.xml
+++ b/pom.xml
@@ -228,8 +228,7 @@
     <version.org.jboss.xnio>3.7.7.Final</version.org.jboss.xnio>
     <version.org.jdom>1.1.3</version.org.jdom>
     <version.org.jfree.jfreechart>1.5.0</version.org.jfree.jfreechart>
-    <version.org.jpmml.model>1.4.11
-    </version.org.jpmml.model> <!-- jpmml-model BSD 3C license - ATTENTION 1.4.11 intentional, because 1.4.9 evaluators depend on 1.4.11 -->
+    <version.org.jpmml.model>1.5.1</version.org.jpmml.model> <!-- jpmml-model BSD 3C license - ATTENTION 1.4.11 intentional, because 1.4.9 evaluators depend on 1.4.11 -->
     <version.org.json>20090211</version.org.json>
     <version.org.jsoup>1.8.3</version.org.jsoup>
     <version.org.junit>5.5.2</version.org.junit>


gitgabrio:

tarilabs:
thanks for this, won't be need an update in the jpmml evaluators in the corresponding repo as well ?

gitgabrio:
> thanks for this, won't be need an update in the jpmml evaluators in the corresponding repo as well ?

Can you point me to that ? What repo it is?

tarilabs:
@gitgabrio 

https://github.com/kiegroup/kie-jpmml-integration/blob/56999e80b220fd21e8668fb549cfb0d6520a629a/pom.xml#L21

to check conformity the same branch name will need be raised from your account

pom.xml
```diff
@@ -228,8 +228,7 @@
     <version.org.jboss.xnio>3.7.7.Final</version.org.jboss.xnio>
     <version.org.jdom>1.1.3</version.org.jdom>
     <version.org.jfree.jfreechart>1.5.0</version.org.jfree.jfreechart>
-    <version.org.jpmml.model>1.4.11
-    </version.org.jpmml.model> <!-- jpmml-model BSD 3C license - ATTENTION 1.4.11 intentional, because 1.4.9 evaluators depend on 1.4.11 -->
+    <version.org.jpmml.model>1.5.1</version.org.jpmml.model> <!-- jpmml-model BSD 3C license - ATTENTION 1.4.11 intentional, because 1.4.9 evaluators depend on 1.4.11 -->
```
manstis, line 231: Did you want to update the comment too?

manstis: .. do you think `because 1.4.9 evaluators depend on 1.4.11` is important?

gitgabrio:
> @gitgabrio
> 
> https://github.com/kiegroup/kie-jpmml-integration/blob/56999e80b220fd21e8668fb549cfb0d6520a629a/pom.xml#L21
> 
> to check conformity the same branch name will need be raised from your account

??????
Anyway, thanks for the link, I'll fix also that - they have to be the same version.

pom.xml
gitgabrio: @manstis 
Yes - even if I do not understand the meaning of it.
I think they should be the same version.
Does anybody of you have any clue why they have different versions ?
@tarilabs ^ ^

tarilabs:
> ??????

I cannot bump the version in the other repo myself, in order for Jenkins CI to check it execute properly the FDB it must be the same branch from the same author.
Since you are the one authoring this PR, you must also be the one in the other repo raising the PR using the same branch name.

> they have to be the same version

Or a compatible one, as mentioned in the comment, executor `1.4.9` evaluators depended on `1.4.11` for version.org.jpmml.model

pom.xml
tarilabs: answered here: https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1374#issuecomment-648843512

tarilabs:
In other words, the evaluators https://mvnrepository.com/artifact/org.jpmml/pmml-evaluator/1.4.9 in the other repository, required `1.4.11` in this repository.

That was odd indeed, and specifically the reason why we took the time to add the comment ;)

tarilabs:
if the jpmml are no longer requiring the odd different version, the comment where it states `"- ATTENTION intentional..."` I believe can just be removed as it does no longer serve the purpose to highlight the numbering scheme was problematic for those versions as explained in https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1374#issuecomment-648845334

gitgabrio:
> if the jpmml are no longer requiring the odd different version, the comment where it states `"- ATTENTION intentional..."` I believe can just be removed as it does no longer serve the purpose to highlight the numbering scheme was problematic for those versions as explained in [#1374 (comment)](https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1374#issuecomment-648845334)

Reading the jpmml-evaluator pom, I'm not 100% sure this is enforced.
I mean - now they are the same, but they are not using a single -static variable; some times ago they where different.

tarilabs:
> I mean - now they are the same, but they are not using a single -static variable; some times ago they where different.

I am not following you, I believe this is exactly what I tried to explain here https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1374#issuecomment-648843512 and here https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1374#issuecomment-648845334 ?

gitgabrio:
@jiripetrlik @manstis @tarilabs 
Could you pls look at this and - eventually - approve? Thx!

PR #1380, opened 2020-07-01
[DROOLS-5464] Add kie-pmml-dependencies version

gitgabrio:
@danielezonca @mariofusco @jiripetrlik

see https://issues.redhat.com/browse/DROOLS-5464

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5292,7 +5292,15 @@
           </exclusion>
         </exclusions>
       </dependency>
+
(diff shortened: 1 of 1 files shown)


gitgabrio:
> @gitgabrio
> I'm not sure about it. Because when I was importing this to my project I had to also add "pom" for it to dependency management. But I'm not 100 % sure what is the correct approach.

You right, my bad...

PR #1382, opened 2020-07-03
remove archiveArtifacts for war and zip files

Ginxo:


Ginxo:
closed because it will fail because container timeout

PR #1389, opened 2020-07-08
makes script fail when optaplanner is not checked out to 7.x branch

mbiarnes:
If optaplanner is checked out to **master** the update-version-all script will fail.
optaplanner has to be checked out to **7.x** when all other reps are checked out to master.
Problem we had: when all reps where checked out to **master** and the script run, optaplanner **master** was updated and lost his version **8.0.0-SNAPSHOT**. 
The optaplanner 7.x should have always the same version as all other **master** branches.

### DIFF
--- a/script/release/update-version-all.sh
+++ b/script/release/update-version-all.sh
@@ -89,6 +89,43 @@ startDateTime=`date +%s`
 
 cd $droolsjbpmOrganizationDir
 
+# --- --- ---
+# checks if optaplanner is on the right 7.x branch if all other repos are checked out to master
+# optaplanner 7.x branch has the same version as master on other kiegroup/reps
+cd optaplanner
+optaBranch=$(git rev-parse --abbrev-ref HEAD)
+if [ $optaBranch == "master" ]; then
+    echo ""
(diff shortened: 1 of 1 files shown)


mareknovotny:
why we should not change the branch to 7.x instead of failing?

mbiarnes:
@mareknovotny also a good option

ge0ffrey:
@mareknovotny autocorrection can be risky. Silently changing the branch can have dire consequences if another command (such as git-all) happened earlier (and the user won't be aware that something special happened). Note that we can **not** make "git-all.sh" automatically change the branch for optaplanner.
In theory, the error never happens, unless the users make a mistake.

I softly recommend the proposed fail fast approach, but this is your team's call and ownership - it won't be impacting me unless the wrong jar ends up on maven central (and we might need to find a good way to sanity check that).

mbiarnes:
@mareknovotny +1 for what @ge0ffrey mentioned someone who is sure he checked out all reps to master and silently the optapalnner 7.x branch is checkout out the user will b surprised as he checked out master

mareknovotny:
I am not really suggesting to silently do that @ge0ffrey @mbiarnes 
There could be an option to fail or fix it for user if he wants, wdyt?

mbiarnes:
@mareknovotny better? the user can choose or to abort or to checkout optaplanner to 7.x

PR #1399, opened 2020-07-16
Downstream

mbiarnes:
@mareknovotny @Ginxo 
The new.downstream.stages were removed and also removed from Jenkinsfile.
We now have 5 Jenkins jobs ending in
.pullrequest, .downstream, .compile, *.downstream.production and upstream.
each job, depending on its ending will call a *.stage.
The downstream.stage we had before this is not any-more used.
Once done this change (and also back-ported to 7.39.x) we should rename the Jenkins jobs and remove the **_new-_** suffix.


downstream.stages
```diff
@@ -1,11 +1,14 @@
...
     configFileProvider([configFile(fileId: SETTINGS_XML_ID, variable: 'MAVEN_SETTINGS_XML_DOWNSTREAM')]) {
         println "Reading file ${REPOSITORY_LIST_FILE}"
         def file = readFile REPOSITORY_LIST_FILE
         def projectCollection = file.readLines()
-        projectCollection.removeAll(['kie-jpmml-integration','droolsjbpm-tools'])
-        treebuild.build(projectCollection, "${SETTINGS_XML_ID}", '-e -nsu -Pbusiness-central,wildfly,sourcemaps,no-showcase clean install -Dfull=true -Dcontainer=wildfly -Dcontainer.profile=wildfly -Dintegration-tests=true -Dmaven.test.failure.ignore=true -Dmaven.test.redirectTestOutputToFile=true -Dgwt.compiler.localWorkers=1 -Dwebdriver.firefox.bin=/opt/tools/firefox-60esr/firefox-bin')
+        def gitURL = env.ghprbAuthorRepoGitUrl ?: env.GIT_URL
+        def project = util.getProjectGroupName(util.getProject(gitURL))[1]
+        projectCollection.removeAll(['droolsjbpm-tools', 'kie-docs'])
+        compile.build(projectCollection, project, "${SETTINGS_XML_ID}", "${PROPERTIES_FILE_ID}")
```
Ginxo, line 12: @mbiarnes why do you call `compile` from the downstream build?

Ginxo: are we using compile for the new downstreams? :thinking:

mbiarnes: @ginxo because we had this from the beginning. All test end everything were done with this. 
We only have in shared libraries or pullrequest.groovy o compile.groovy, there is not (yet) existing a downstream.groovy.
Should it be treebuild.groovy?

mbiarnes:
Please merge this after #1401

mbiarnes:
@Ginxo after reviewing we had from the beginning of using these new files the compile.groovy in jenkins-pipeline-shared-libraries for FDBs and CDBs

mbiarnes:
OK - I will close this and reopen a new one

PR #1401, opened 2020-07-17
BXMSPROD-878: moved properties files used by getGoals from Jenkins

mbiarnes:
https://issues.redhat.com/browse/BXMSPROD-878
I hope everything what had to was changed
please close https://github.com/kiegroup/appformer/pull/1003 after merging
depends on https://github.com/kiegroup/jenkins-pipeline-shared-libraries/pull/87


mareknovotny:
what if we externalize some of the maven argument commands and JVM options to `.mvn/maven.config` and `.mvn/jvm.config`
see https://maven.apache.org/configure.html

We don't need to do that now and for all repositories, just an idea.

mbiarnes:
Please merge together with: https://github.com/kiegroup/jenkins-pipeline-shared-libraries/pull/86

mbiarnes:
the test I did if this works are:
PR: https://rhba-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/custom/job/mbiarnes/job/mbiarnes-appformer.pullrequest/7/console
FDB: https://rhba-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/custom/job/mbiarnes/job/mbiarnes-appformer.downstream/1/console
CDB: https://rhba-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/custom/job/mbiarnes/job/mbiarnes_appformer.compile/1/console

.jenkins/compilation.stages
```diff
@@ -10,6 +10,9 @@ stage('Compile Downstream Build') {
         def gitURL = env.ghprbAuthorRepoGitUrl ?: env.GIT_URL
         def project = util.getProjectGroupName(util.getProject(gitURL))[1]
         projectCollection.removeAll(['droolsjbpm-tools', 'kie-docs'])
-        compile.build(projectCollection, project, "${SETTINGS_XML_ID}", "${PROPERTIES_FILE_ID}")
+        def propContent = readFile PROPERTIES_FILE_PATH
```
almope, line 13: I would consider moving this logic to the Jenkins shared library either into the `getGoals` function or even better as another function set like `propertiesFromPath(...)`, `propertiesFromFileId(...)`, etc., and use that to supply the required parameters from here.

.gitignore
```diff
@@ -3,6 +3,7 @@
 
 # Eclipse, Netbeans and IntelliJ files
 /.*
+!.jenkins
```
almope, line 6: A very minor thing here, if this is a legitimate part of the project, I wouldn't hide it. Also I would use a generic name for the folder, like for example `ci`. Anyway is a good move to stop polluting the root folder of the project :)

Ginxo: probably `.ci` is a more generic one and can be used for future purposes not related just with Jenkins, so @almope is probably right

.jenkins/compilation.stages
Ginxo: In case we do it we should pass absolute path to it, something like `"${env.WORKSPACE}/.jenkins/compileDownstreamBuild.properties`. `util` script would be the proper place for this.
There is one problem here. The `.jenkins` (or `.ci`) folder can be `"${env.WORKSPACE}/.jenkins/compileDownstreamBuild.properties` or `"${env.WORKSPACE}/droolsjbpm-build-bootstrap/.jenkins/compileDownstreamBuild.properties`

- https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/aa2723e865b8bb0b315827f5e9e5083c39ab9aa3/Jenkinsfile#L49
- https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/aa2723e865b8bb0b315827f5e9e5083c39ab9aa3/Jenkinsfile#L59

In this case we should wrap stage files content in a method in order to pass folder as an argument
```
def execute(def ciFolderPath)
stage('New Downstream Build') {
...
...
...
final PROPERTIES_FILE_PATH = "${ciFolderPath}/compileDownstreamBuild.properties"
...
...
...
}
```
and change load calling in case it is in `WORKSPACE` folder https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/aa2723e865b8bb0b315827f5e9e5083c39ab9aa3/Jenkinsfile#L49
```
def stage = load("$WORKSPACE/${file}")
stage.execute(''$WORKSPACE/.jenkins")

```
and in case it is `droolsjbpm-build-bootstrap` https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/aa2723e865b8bb0b315827f5e9e5083c39ab9aa3/Jenkinsfile#L59
```
def stage = load("${file}")
stage.execute(''$WORKSPACE/droolsjbpm-build-bootstrap/.jenkins")
```

mbiarnes:
@Ginxo @almope All proposed changes are implemented

mbiarnes:
LGTM

mbiarnes:
@Ginxo finally we go with .ci folder?

Ginxo:
@mbiarnes I think it's more aligned with `.git` and `.github` folders approach, since the dot more or less means it's not part of the source code itself

PR #1425, opened 2020-08-05
Update eclipse-code-style-formatter_droolsjbpm-java-conventions.xml

fjtirado:
I think 220 is too large, 120 is aligned with Intellij one


mareknovotny:
> I think 220 is too large, 120 is aligned with Intellij one
do you mean in line with default one in Intellij, right? as here https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/ide-configuration/intellij-configuration/code-style/intellij-code-style_droolsjbpm-java-conventions.xml doesn't have specified wrapping so it takes the default one.
Eclipse's default one is also 120 letters.
I do not have any objection to this change.

mareknovotny:
so tagging another eclipse user in our team to approve it.

PR #1432, opened 2020-08-11
Update version

mbiarnes:
this fixes the update procedure
(try it locally first)

### DIFF
--- a/kie-dmn-bom/pom.xml
+++ b/kie-dmn-bom/pom.xml
@@ -10,7 +10,6 @@
     <relativePath>../kie-user-bom-parent/pom.xml</relativePath>
   </parent>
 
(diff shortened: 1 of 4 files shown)


kie-build-tools/pom.xml
```diff
@@ -10,7 +10,9 @@
     <relativePath>../kie-user-bom-parent/pom.xml</relativePath>
   </parent>
 
+  <groupId>org.kie</groupId>
   <artifactId>kie-build-tools</artifactId>
+  <packaging>pom</packaging>
```
winklerm, line 15: Is this really correct? The module packages findbugs resources, not sure if using only pom would not exclude them?

mbiarnes: removed - was wrong!

PR #59, opened 2015-04-24
Changes for Initial GWT 2.7.0 setup

porcelli:

PR #1442, opened 2020-08-20
AF-2542: GWT 2.9 upgrade

porcelli:
**JIRA**: [AF-2542](https://issues.redhat.com/browse/AF-2542)

**referenced Pull Requests**: 

* https://github.com/kiegroup/lienzo-core/pull/122
* https://github.com/kiegroup/lienzo-tests/pull/103
* https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1442
* https://github.com/kiegroup/appformer/pull/1030
* https://github.com/kiegroup/drools/pull/3051
* https://github.com/kiegroup/kie-wb-common/pull/3395
* https://github.com/kiegroup/drools-wb/pull/1412

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -316,7 +316,7 @@
     <version.org.jboss.arquillian.extension.drone>2.5.1</version.org.jboss.arquillian.extension.drone>
     <version.org.jboss.arquillian.graphene>2.3.2</version.org.jboss.arquillian.graphene>
     <version.com.google.guava.for-selenium>23.0</version.com.google.guava.for-selenium>
-    <version.com.google.elemental2>1.0.0-RC1</version.com.google.elemental2>
+    <version.com.google.elemental2>1.1.0</version.com.google.elemental2>
     <version.org.jboss.errai>4.9.0-SNAPSHOT</version.org.jboss.errai>
     <!-- WildFly version used together with  the GWT's Super Dev Mode -->
     <version.org.jboss.errai.wildfly>19.1.0.Final</version.org.jboss.errai.wildfly>
--- a/pom.xml
+++ b/pom.xml
@@ -54,7 +54,7 @@
       - Version properties must be sorted alphabetically (other form of sorting were found to be unclear and ambiguous).
     -->
     <version.aopalliance>1.0</version.aopalliance>
-    <version.asm>7.2</version.asm>
+    <version.asm>7.1</version.asm>
     <version.ch.qos.cal10n>0.8.1</version.ch.qos.cal10n>
     <version.ch.qos.logback>1.2.3</version.ch.qos.logback>
     <version.commons-beanutils>1.9.4</version.commons-beanutils>
@@ -77,10 +77,10 @@
     <version.com.fasterxml.jackson>2.10.4</version.com.fasterxml.jackson>
     <version.com.fasterxml.jackson.databind>2.10.4</version.com.fasterxml.jackson.databind>
     <version.com.fasterxml.jackson.annotations>2.9.10</version.com.fasterxml.jackson.annotations>
-    <version.com.google.code.gson>2.8.2</version.com.google.code.gson>
+    <version.com.google.code.gson>2.8.6</version.com.google.code.gson>
     <version.com.google.guava>25.0-jre</version.com.google.guava>
     <version.com.google.inject.guice>4.0</version.com.google.inject.guice>
-    <version.com.google.gwt>2.8.2</version.com.google.gwt>
+    <version.com.google.gwt>2.9.0</version.com.google.gwt>
     <version.org.google.gwt-charts>0.9.9</version.org.google.gwt-charts>
     <version.com.google.gwt.gwtmockito>1.1.9</version.com.google.gwt.gwtmockito>
     <version.com.google.jsinterop.base>1.0.0</version.com.google.jsinterop.base>
@@ -317,7 +317,7 @@
     <version.org.jboss.arquillian.graphene>2.3.2</version.org.jboss.arquillian.graphene>
     <version.com.google.guava.for-selenium>23.0</version.com.google.guava.for-selenium>
     <version.com.google.elemental2>1.0.0-RC1</version.com.google.elemental2>
-    <version.org.jboss.errai>4.8.0.Final</version.org.jboss.errai>
+    <version.org.jboss.errai>4.9.0-SNAPSHOT</version.org.jboss.errai>
     <!-- WildFly version used together with  the GWT's Super Dev Mode -->
     <version.org.jboss.errai.wildfly>19.1.0.SP1</version.org.jboss.errai.wildfly>
     <!-- Required by jboss-remoting version above -->
--- a/pom.xml
+++ b/pom.xml
@@ -5308,18 +5308,21 @@
         <scope>test</scope>
       </dependency>
       
-      <!-- Used by jbpm-workitems-rest -->
+      <!-- Used by jgit -->
       <dependency>
         <groupId>org.bouncycastle</groupId>
         <artifactId>bcprov-jdk15on</artifactId>
         <version>${version.org.bouncycastle}</version>
-        <scope>test</scope>
       </dependency>
       <dependency>
         <groupId>org.bouncycastle</groupId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -170,9 +170,9 @@
     <version.org.eclipse.aether>1.1.0</version.org.eclipse.aether>
     <version.org.eclipse.emf>2.6.0.v20100614-1136</version.org.eclipse.emf>
     <version.org.eclipse.emf.ecore.xmi>2.5.0.v20100521-1846</version.org.eclipse.emf.ecore.xmi>
-    <version.org.eclipse.jdt>3.19.0</version.org.eclipse.jdt>
+    <version.org.eclipse.jdt>3.18.0</version.org.eclipse.jdt>
```
hasys, line 173: Hi @porcelli, why jdt was downgraded?

pom.xml
```diff
@@ -54,7 +54,7 @@
       - Version properties must be sorted alphabetically (other form of sorting were found to be unclear and ambiguous).
     -->
     <version.aopalliance>1.0</version.aopalliance>
-    <version.asm>3.3.1</version.asm>
+    <version.asm>7.1</version.asm>
```
danielezonca, line 57: This version doesn't exist: after `3.3.1` they changed the groupdId to `org.ow2.asm` and the corresponding property is `version.org.ow2.asm` that is already `7.1`

Btw we should probably review the usages of `asm:asm:3.3.1` and try to migrate to the newer version but this is a comment for another ticket :)

pom.xml
danielezonca, line 173: As far as I remember GWT (`gwt-dev`) include its own version of `jdt` directly in the jar and this version is only used by `drools-compiler`.
To avoid classpath clashing and similar, we shaded the version used by Drools ( [link](https://github.com/kiegroup/drools/blob/c7fba5bd3ffdacadfea2eabfb00e47467c463d94/drools-compiler/pom.xml#L207-L235) ) and after that change we had no other `jdt` upgrade problem like in the past.

I just noticed that together with Errai GWT 2.9 upgrade, it has been added an explicit dependency to `jdt` ( [link](https://github.com/errai/errai/commit/0ecd8fd5ed1920f8357a4d9989f968d79b429a9c) ) for example in `errai-codegen`. Is it possible in Errai to continue to use `jdt` provided by GWT? Or if not possible what about shade it so that it will not leak here?

We need to preserve the possibility to upgrade `jdt` for drools core when needed and this downgrade might break other stuff on Drools (and Kogito too).

/cc @mariofusco

danielezonca: Ok I just noticed we are already shading it in errai ( [link](https://github.com/errai/errai/blob/master/errai-codegen/pom.xml#L142) ) so my only question is why this downgrade? :)

danielezonca:
jenkins execute fdb

pom.xml
```diff
@@ -170,9 +170,9 @@
     <version.org.eclipse.aether>1.1.0</version.org.eclipse.aether>
     <version.org.eclipse.emf>2.6.0.v20100614-1136</version.org.eclipse.emf>
     <version.org.eclipse.emf.ecore.xmi>2.5.0.v20100521-1846</version.org.eclipse.emf.ecore.xmi>
-    <version.org.eclipse.jdt>3.19.0</version.org.eclipse.jdt>
+    <version.org.eclipse.jdt>3.18.0</version.org.eclipse.jdt>
     <version.org.eclipse.jetty>9.2.14.v20151106</version.org.eclipse.jetty>
-    <version.org.eclipse.jgit>5.0.2.201807311906-r</version.org.eclipse.jgit>
+    <version.org.eclipse.jgit>5.7.0.202003110725-r</version.org.eclipse.jgit>
```
mareknovotny, line 175: Is this version upgrade required by GWT 2.9 upgrade or by something else?

pom.xml
```diff
@@ -313,10 +313,10 @@
     <version.org.jboss.arquillian.extension.drone>2.5.1</version.org.jboss.arquillian.extension.drone>
     <version.org.jboss.arquillian.graphene>2.3.2</version.org.jboss.arquillian.graphene>
     <version.com.google.guava.for-selenium>23.0</version.com.google.guava.for-selenium>
-    <version.com.google.elemental2>1.0.0-RC1</version.com.google.elemental2>
-    <version.org.jboss.errai>4.8.0.Final</version.org.jboss.errai>
+    <version.com.google.elemental2>1.1.0</version.com.google.elemental2>
+    <version.org.jboss.errai>4.9.0.Final</version.org.jboss.errai>
     <!-- WildFly version used together with  the GWT's Super Dev Mode -->
-    <version.org.jboss.errai.wildfly>19.1.0.SP1</version.org.jboss.errai.wildfly>
+    <version.org.jboss.errai.wildfly>19.1.0.Final</version.org.jboss.errai.wildfly>
```
mareknovotny, line 319: This is not correct. we should keep 19.1.0.SP1 version as the 19.1.0.Final wrapped Errai WF distro has wrong execute permission in zip file

mareknovotny (changes requested):
There is wrong Errai WF distro version change

pom.xml
porcelli: to align with GWT version

porcelli: align dependencies creates less problems, and the downgrade had no affect on runtimes

pom.xml
porcelli: i don't recall exactly what GWT related dependency is associated it with... but this bump is related to GWT upgrade indeed.

pom.xml
danielezonca: I'm fine with this, my only concern is to avoid if possible to have to keep them aligned: we already had in the past the need to upgrade `ecj` for `drools-compiler` and I would like to preserve this possibility. Do you expect this upgrade will prevent us/increase the complexity to have different versions for GWT/drools-compiler in the future? If this is and will be possible I'm fine with this downgrade :)

porcelli: I'd expect this would be the last upgrade related to GWT in this stream, so keep it simple for now... won't prevent us to do better in future.

But for now I'd prefer the simplification :)

mareknovotny:
Jenkins do fdb

mbiarnes:
Jenkins do fdb

barboras7:
Jenkins do fdb

mareknovotny:
@barboras7 please don't rerun FDB again and again this is clear it won't pass as Jenkins doesn't take correct branches from @porcelli forks which doesn'ŧ match his github account name

mareknovotny:
launching FDB after https://github.com/kiegroup/jenkins-pipeline-shared-libraries/pull/100 is merged

Jenkins do fdb

porcelli:
failing tests seem to be completely unrelated (one heap and 2 dependencies issues on jbpm runtime side)

mareknovotny:
@porcelli yes, the failures are not related, but the FDB didn't finished due timeout, so not all repositories were built. I am trying to manually increase timeout and replay the #59 build with https://rhba-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/KIE/job/master/job/fullDownstream/job/new-FDB-droolsjbpm-build-bootstrap-master.downstream/61/

porcelli:
Jenkins do fdb

porcelli:
the failing test seems to be unrelated =/

porcelli:
Jenkins do fdb

mareknovotny:
Jenkins do fdb

mareknovotny:
last FDB run shown the not enough memory for Arquillian test org.kie.workbench.common.screens.server.management.backend.StandaloneControllerIT.org.kie.workbench.common.screens.server.management.backend.StandaloneControllerIT 
```
org.jboss.arquillian.container.spi.client.container.DeploymentException: Cannot deploy: workbench.war
Caused by: java.lang.OutOfMemoryError: Java heap space
```

mareknovotny:
Could we close this PR set if there is active PR set https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1488 @porcelli ?

mareknovotny:
closing as there is ongoing work on https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1488

PR #1465, opened 2020-09-15
[BAQE-1547] git-clone-others.sh - adding upstream remote branch should always be done + new update kie old version script

afalhambra:
**[BAQE-1547](https://issues.redhat.com/browse/BAQE-1547)**: 
- git-clone-others.sh - adding upstream remote branch should always be done
- New script added to update old kie version property automatically

### DIFF
--- a/script/git-clone-others.sh
+++ b/script/git-clone-others.sh
@@ -89,6 +89,7 @@ done
 
 
 for repository in $REPOSITORY_LIST ; do
+    gitUrlPrefix=${droolsjbpmGitUrlPrefix}
     echo
     if [ -d $repository ] ; then
         echo "==============================================================================="
@@ -98,7 +99,6 @@ for repository in $REPOSITORY_LIST ; do
         echo "==============================================================================="
         echo "Repository: $repository"
         echo "==============================================================================="
-        gitUrlPrefix=${droolsjbpmGitUrlPrefix}
         echo -- prefix ${gitUrlPrefix} --
         echo -- repository ${repository} --
         echo -- ${gitUrlPrefix}${repository}.git -- ${repository} --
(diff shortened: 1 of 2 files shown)


script/git-clone-others.sh
```diff
@@ -89,6 +89,7 @@ done
 
 
 for repository in $REPOSITORY_LIST ; do
+    gitUrlPrefix=${droolsjbpmGitUrlPrefix}
```
mbiarnes, line 92: @afalhambra why this was taken out if the second if statement?

mbiarnes:
@afalhambra this scripts worked for ages - why is it better to change it this way?

script/git-clone-others.sh
afalhambra: to make it available for all the different branches in the if conditions below, where this var can be used.
For instance:

https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/2c2fdf86e2c3cd1488c26b05ec89449fff3438e0/script/git-clone-others.sh#L128

and

https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/2c2fdf86e2c3cd1488c26b05ec89449fff3438e0/script/git-clone-others.sh#L120

afalhambra:
> @afalhambra this scripts worked for ages - why is it better to change it this way?

The reason for this change is because in case the repos are already in the filesystem, the argument passed to add remote upstream `--add-upstream-remote` to the existing repos won't be executed. I'm just getting this out of the `if statement` cause my understanding is that it should add this upstream regardless of whether the repos are in the filesystem or not.

It's just a minor improvement, if you think it doesn't add any further value, I can discard this change. No problems with that

MarianMacik (approved):
I think it looks good and makes sense to me. Wdyt @mbiarnes ?

mbiarnes (approved):
+1

PR #1474, opened 2020-09-21
[DROOLS-5367] Updated kie-pmml bom

gitgabrio:
@danielezonca @mariofusco @jiripetrlik

See https://issues.redhat.com/browse/DROOLS-5367

This PR just add a new module to kie-pmml bom

### DIFF
--- a/kie-pmml-bom/pom.xml
+++ b/kie-pmml-bom/pom.xml
@@ -205,7 +205,6 @@
       <dependency>
         <groupId>org.kie</groupId>
         <artifactId>kie-pmml-models-tests</artifactId>
-        <classifier>tests</classifier>
         <version>${version.org.kie}</version>
       </dependency>
       <!-- Regression -->
--- a/kie-pmml-bom/pom.xml
+++ b/kie-pmml-bom/pom.xml
@@ -174,6 +174,7 @@
       </dependency>
       <!-- EXTERNAL -->
       <!-- TEST -->
+      <!-- Commons -->
       <dependency>
         <groupId>org.kie</groupId>
         <artifactId>kie-pmml-commons</artifactId>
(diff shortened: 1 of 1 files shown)


danielezonca (approved):

kie-pmml-bom/pom.xml
```diff
@@ -193,6 +194,19 @@
...
         <version>${version.org.kie}</version>
       </dependency>
+      <!-- Evaluator -->
+      <dependency>
+        <groupId>org.kie</groupId>
+        <artifactId>kie-pmml-evaluator-core</artifactId>
+        <classifier>tests</classifier>
+        <version>${version.org.kie}</version>
+      </dependency>
+      <!-- Tests -->
```
jiripetrlik, line 204: @gitgabrio @mareknovotny I'm not convinced about this one. Do we want to add versions of test modules into bom for users?

jiripetrlik (commented):
I have two questions about this:
1. First is about adding test dependencies to boms for users
2. I'm not sure if this PR is related to DROOLS-5367. It was about avoiding of kmodule.xml and not about update of bom.

kie-pmml-bom/pom.xml
mareknovotny: it depends if this is for end users to consume in their applications.
@gitgabrio this and below GAV for internal testing or for end user consumption ?
you can have it added in kie-parent if that is only internal dependencyManagement

gitgabrio: @mareknovotny @jiripetrlik @danielezonca 
One of the reason behind the architecture I designed for PMML it is to allow the "end users" to write their own model implementation(s), overriding/replacing our(s).
From this point of view, even this module should be considered available for end users.
Does this make sense ?

gitgabrio: @danielezonca @jiripetrlik 
Hi! Could we proceed with this one? It is needed by other ones currently opened and one way or another I would like to close it

mareknovotny:
Please rebase the https://github.com/kiegroup/drools/pull/3117 and we should get some feedback from @jiripetrlik to your last explanations why it is added into kie-pmml-bom

kie-pmml-bom/pom.xml
danielezonca: Fine for me to expose this module if you expect other user can need it

gitgabrio:
@mareknovotny 
> Please rebase the [kiegroup/drools#3117](https://github.com/kiegroup/drools/pull/3117)

Done

kie-pmml-bom/pom.xml
jiripetrlik: @gitgabrio Understand, I'm fine with merging this.

jiripetrlik (approved):

PR #1481, opened 2020-09-30
Aligning eclipse formatter with intellij one

fjtirado:
Although diff shows a lot of changes, I just modify the wrapping setting for method invocations to use indentations rather than tabular form


ricardozanini (approved):
I assume the diff is just a matter of formatting thing.

cristianonicolai (approved):

PR #1496, opened 2020-10-09
Eclipse format change

fjtirado:


cristianonicolai (approved):

mareknovotny:
I noticed this by accident, could you announce it for other eclipse users too and to reload their settings ?

PR #1499, opened 2020-10-13
[DROOLS-5688] Declaring trusty-pmml public API

gitgabrio:
@danielezonca @mareknovotny  @jiripetrlik

See https://issues.redhat.com/browse/DROOLS-5688

This PR declare the version of org.kie/kie-pmml-api

Needed by https://github.com/kiegroup/drools/pull/3162

### DIFF
--- a/kie-pmml-bom/pom.xml
+++ b/kie-pmml-bom/pom.xml
@@ -67,6 +67,11 @@
       <!-- PMML -->
       <dependency>
         <groupId>org.kie</groupId>
+        <artifactId>kie-pmml-api</artifactId>
+        <version>${version.org.kie}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.kie</groupId>
         <artifactId>kie-pmml-commons</artifactId>
         <version>${version.org.kie}</version>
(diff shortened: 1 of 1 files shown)


kie-pmml-bom/pom.xml
```diff
@@ -65,6 +65,11 @@
   <dependencyManagement>
     <dependencies>
       <!-- PMML -->
+      <dependency>
+        <groupId>org.kie</groupId>
+        <artifactId>kie-pmml-api</artifactId>
+        <version>${version.org.kie}</version>
```
danielezonca, line 71: Can you please duplicate this dependency (and all the other of the file) with `<classifier>sources</classifier>`?

danielezonca (approved):

kie-pmml-bom/pom.xml
gitgabrio: @danielezonca 
What's the reason for that ? Duplicating all of them would make maintaining harder, I would prefer to avoid if not really needed/requested

jiripetrlik (approved):

PR #1488, opened 2020-10-01
AF-1799: Upgrade JDK 11 + GWT 2.9

caponetto:
**Upgrade JDK 11 + GWT 2.9**

Feel free to comment and/or raise any concerns.

This PR continues the work done in 
- https://github.com/kiegroup/appformer/pull/1030
- https://github.com/kiegroup/appformer/pull/1039

Thanks guys for that!

**JIRA**:

[AF-1799](https://issues.redhat.com/browse/AF-1799)

**Parent Pull Request**:
- https://github.com/kiegroup/lienzo-core/pull/134

<details>
<summary>
How to retest this PR or trigger a specific build:
</summary>

* <b>a pull request</b> please add comment: <b>Jenkins retest this</b>
 
* <b>a full downstream build</b> please add comment: <b>Jenkins run fdb</b>
  
* <b>a compile downstream build</b> please  add comment: <b>Jenkins run cdb</b>

* <b>a full production downstream build</b> please add comment: <b>Jenkins execute product fdb</b>

* <b>an upstream build</b> please add comment: <b>Jenkins run upstream</b>
</details>

PR #1553, opened 2020-12-09
errai upgrade to 4.11.0.Final

mbiarnes:
errai upgrade


tiagobento:
@mbiarnes Actually, I think we can do this upgrade as part of the JDK11/GWT2.9 branches. We don't want Errai to be upgraded so close to the cut-off. My recommendation is to close this PR and upgrade Errai on #1488.

tiagobento:
@caponetto ^

mbiarnes:
this PR will be closed since the errai upgrade to 4.11.0.Final will be done #1488

PR #1742, opened 2021-08-30
Add a specific property for jakarta-el dependency version

almope:
Signed-off-by: Alberto Morales Perez <almorale@redhat.com>


almope:
Jenkins execute fdb

mareknovotny (commented):
looks good like we discussed and agreed as fast fix for having different versions of el-api and el impl

PR #1756, opened 2021-09-07
BXMSPROD-1468: skipping to build PIM when using jdk1.8

mbiarnes:
**Thank you for submitting this pull request**

**JIRA**: [BXMSPROD-1468](https://issues.redhat.com/browse/BXMSPROD-1468)

- when the reps are build with >= jdk11 the jars will be compiled the jdk1.8 could read them
- when the JAVA version for compiling the reps is **jdk 1.8** the mvn-all.sh scripts skips PIM,  removing this project from repository-list.txt 

Works locally. 
mvn-all.sh was used with jdk1.8 and jdk11, it worked.
In case the build was with jdk 11, a *.jar was downloaded, a class extracted and 
`javap -v XXX.class | grep major` done with 52 (for jdk1.8) as result.

<details>
<summary>
How to retest this PR or trigger a specific build:
</summary>

* <b>a pull request</b> please add comment: <b>Jenkins retest</b> (using <i>this</i> e.g. <b>Jenkins retest this</b> optional but no longer required)
 
* <b>a full downstream build</b> please add comment: <b>Jenkins run fdb</b>
  
* <b>a compile downstream build</b> please  add comment: <b>Jenkins run cdb</b>

* <b>a full production downstream build</b> please add comment: <b>Jenkins execute product fdb</b>

* <b>an upstream build</b> please add comment: <b>Jenkins run upstream</b>
</details>

PR #1761, opened 2021-09-08
Switch nightly build to JDK11

almope:
Signed-off-by: Alberto Morales Perez <almorale@redhat.com>

### DIFF
--- a/.ci/jenkins/Jenkinsfile.nightly
+++ b/.ci/jenkins/Jenkinsfile.nightly
@@ -6,7 +6,7 @@ pipeline {
     }
     tools {
         maven 'kie-maven-3.8.1'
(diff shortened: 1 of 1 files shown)


almope:
Jenkins execute fdb

almope:
Jenkins do fdb

mareknovotny:
this only partial switch, right? we still need to use `maven.compiler.release` parameter to be 8

almope:
@mareknovotny It's a full switch to JDK 11 compilation but with the `maven.compiler.release` property, which was merged as part of this PR: https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1756

mareknovotny:
ok, i missed that

PR #1799, opened 2021-11-02
composite actions

Ginxo:
related PRs:

- https://github.com/kiegroup/lienzo-core/pull/157
- https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1799

**Pros**:

- the common steps are centralized
- one single place to upgrade build-chain versioning -> one single PR everytime we want to upgrade
- since the definition-file input for build-chain/action.yml has a default value the definition-file can be also centralized

**Cons**:

- Not possible to test action changes (unless we temporary point to the open PR repo and branch, as it is done for this one like `ginxo/droolsjbpm-build-bootstrap/.ci/actions/java@tests/composite-actions`)
- additional complexity level

Personal opinion: "I would go for it"


winklerm (commented):
@Ginxo I am not a GHA expert, but I like it very much - it helps the maintainability and even though it brings a bit of complexity, I think it is totally worth it.

almope:
I agree with @winklerm. My only concern is whether we will end up needing different versions for the same action and different projects. I guess it's possible to manage this with the `input` parameters. Overall I think it's a good improvement.

almope (approved):
Thanks!

Ginxo:
> I agree with @winklerm. My only concern is whether we will end up needing different versions for the same action and different projects. I guess it's possible to manage this with the `input` parameters. Overall I think it's a good improvement.

@almope good point. In case this is needed it will be about branching repository and to point to `kiegroup/droolsjbpm-build-bootstrap/.ci/actions/java@v1` or `kiegroup/droolsjbpm-build-bootstrap/.ci/actions/java@v2`  (then we will think about to externalize composite actions to an external repository (like rhba-pipelines or whatever))

danielezonca:
Well done, I'm happy to see this happening :) 

> **Pros**:
> 
> * the common steps are centralized
> * one single place to upgrade build-chain versioning -> one single PR everytime we want to upgrade

Just to be sure, I think you are proposing to make all repos refer to latest commit of `droolsjbpm-build-bootstrap` right?
Something like 
```yaml
uses: kiegroup/droolsjbpm-build-bootstrap/.ci/actions/java@main
```
(or `@latest` I'm not sure about how this work)

> * since the definition-file input for build-chain/action.yml has a default value the definition-file can be also centralized
> 
> **Cons**:
> 
> * Not possible to test action changes (unless we temporary point to the open PR repo and branch, as it is done for this one like `ginxo/droolsjbpm-build-bootstrap/.ci/actions/java@tests/composite-actions`)

Personally I don't think this is a real con, at the moment is sort of the same because the change needs to be manually propagated to each repo with a different PR

> * additional complexity level

I think one additional con is that it makes the build a bit less "reproducible" because it downloads the latest version of the composite step so if I rerun a build it could take a different version of it.
Btw this is already like that because build-chain action configuration is already externalized to a [different repo](https://github.com/kiegroup/droolsjbpm-build-bootstrap/tree/main/.ci).

> 
> Personal opinion: "I would go for it"

Yes I think we should apply this change to all the repos and also to the `kogito-pipeline` setup 👍

Ginxo:
@almope @winklerm @danielezonca I've created an epic for this. Closing test PR
https://issues.redhat.com/browse/BXMSPROD-1538

PR #1846, opened 2021-12-13
remove log4j version declaration

mariofusco:

PR #1847, opened 2021-12-13
upgrade to log4j 2.15.0

mariofusco:


mariofusco:
jenkins execute fdb

mareknovotny:
i would instead use https://github.com/kiegroup/appformer/pull/1218 and https://github.com/kiegroup/kie-soup/pull/239 
No need to upgrade if we don't use it.

mariofusco:
@mareknovotny I agree, but if we're not using this dependency we could at least merge this https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1846

PR #2222, opened 2023-02-15
Fix directory name in which we want to run sonar analysis

shubhbapna:
The naming convention of the cloned folder is `<owner>_<repository>`. So for example instead of `kiegroup_droolsjbpm_integration` we need `kiegroup_droolsjbpm-integration`


lampajr (approved):

Ginxo (approved):
LGTM

PR #2283, opened 2023-05-03
remove unused matrix from generate_status_page.yaml

Ginxo:


lampajr (changes requested):
I think you can safely remove the whole `strategy` field as we don't need `fail-fast` option since there is a single run, wdyt?

lampajr (approved):

PR #2440, opened 2024-03-13
[incubator-kie-issues-994] add in memory token logins

elguardian:
issue: https://github.com/apache/incubator-kie-issues/issues/994


mareknovotny:
jenkins do fdb

PR #2550, opened 2025-01-30
[kie-issues#1788] Updating keycloak to 26.0.1

LightGuard:
Fixes https://github.com/apache/incubator-kie-issues/issues/1788

I think this will cover everything if the whole build is run, but another set of eyes on the whole dependency graph would be useful.

This syncs the version of Keycloak with what Quarkus 3.15.3 is using


LightGuard:
jenkins do fdb

baldimir (approved):

